PR #2, opened 2025-09-12
Refactor Project structure and add devcontainer

JPVenson:
- Add .devcontainer definition for easy development
- Refactor project stucture to allow vscode IDE to actually provide intellisense


src/Anthropic/Client/Anthropic.Client.csproj
```diff
@@ -24,15 +24,13 @@
     <Configurations>Debug;Release</Configurations>
```
sd-st: FYI this will be overwritten as `dotnet csharpier format .` causes this file to be formatted the way it is

src/Anthropic/Anthropic.sln
```diff
@@ -0,0 +1,36 @@
+
```
sd-st: We're actually replacing the `Anthropic/Client` structure with `Anthropic.Client` so I think we probably don't need to move this file anymore https://github.com/anthropics/anthropic-sdk-csharp/pull/1
Correct me if I'm wrong though!

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,9 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
```
sd-st, line 1: Hmm, I'm personally a little wary about putting `devcontainer` or editor-specific stuff in our C# SDKs. It's not something we generally do for our other SDKs, plus I couldn't find these in a few non-Stainless C# SDKs I looked at. I'm primarily worried about needing to maintain these files, plus the fact that we're implicitly endorsing these projects. (There's also the whole supply-chain issue with bringing in external dependencies)

I'm going to see what the team thinks, just wanted to explain my thinking a bit

src/Anthropic/Anthropic.sln
JPVenson: The solution should include all projects that are below it, the old one does not include all of them so i would rather opt for moving the solution anyway

.devcontainer/devcontainer.json
JPVenson: the devcontainer argume i can somewhat understand, its my prefered development enviorment for stuff like this for quite some time now it also allows us to set the pre-requisites directly. For the .vscode configuration, they contain all requirements to work with this code in vscode which is a very widely used tool and it helps tremendously for anyone that just wants to checkout the code and have it working. Its also essential for us to have a single development enviorment and not having the whole dependency hell where one is missing a plugin and cannot work with the project like the rest.

src/Anthropic/Anthropic.sln
sd-st: I think I'm a bit confused, in #1 we end up with `src/Anthropic.Client` and `src/Anthropic.Client.Tests`, and nothing else, and both of them are in the top-level solution

.devcontainer/devcontainer.json
sd-st: I spoke with the team and it looks like we actually do create `.devcontainer` for other languages, so that part is probably fine.

For VSCode, we're just giving them extensions to install aren't we? Won't most C# devs already know what extensions they want? Are there extensions we really need that most C# devs won't have already?

JPVenson: yea but thats how devcontainers work too, you get a blank enviorment and some extensions like the c# language server need to run in the devcontainer not the vscode client so even if you have them setup in your client vscode instance it will tell you:

image

Which you then have to do every time again

RobertCraigie: fwiw in other SDKs we just inline the extensions in the devcontainer config, I don't feel strongly either way tbh

PR #6, opened 2025-10-29
Implement message Aggregator

JPVenson:
- Added the logfile to gitignore
- Adapted devcontainer to install node
- Added Streaming aggregators for both message endpoints


scripts/test
```diff
@@ -53,4 +53,4 @@ else
 fi
 
 echo "==> Running tests"
-dotnet test
+dotnet test src/Anthropic
```
sd-st: Curious, why was this needed? I've been running `scripts/test` without this and it's worked fine; were you having issues?

.devcontainer/devcontainer.json
```diff
@@ -4,5 +4,8 @@
     "name": "C# (.NET)",
     "image": "mcr.microsoft.com/devcontainers/dotnet:1-9.0-noble",
     "postAttachCommand": "cat .vscode/extensions.json | jq -r .recommendations[] | xargs -n 1 code --install-extension",
-    "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}"
-}
+    "workspaceFolder": "/workspaces/${localWorkspaceFolderBasename}",
+    "features": {
+        "ghcr.io/devcontainers/features/node:1": {},
```
sd-st, line 10: Is this for prisma?

scripts/test
JPVenson: for me it complained that it could not find any test.

.devcontainer/devcontainer.json
JPVenson: yea i wasn't able to run mock/test scripts without node and its not installed by default

scripts/test
sd-st: Do you mind pasting the output? I don't think Tomer or I have run into this; would be good to reproduce it on our end

src/Anthropic.Client.Tests/Services/Messages/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
+        // arrange
+        var messagesServiceMock = new Mock<IMessageService>();
+        static async IAsyncEnumerable<RawMessageStreamEvent> GetTestValues()
+        {
+            yield return new(new RawMessageStartEvent(GenerateStartMessage()));
+            yield return new(new RawMessageStopEvent());
+            await Task.CompletedTask;
+        }
+        // act
+        messagesServiceMock.Setup(e => e.CreateStreaming(It.IsAny<MessageCreateParams>())).Returns(GetTestValues);
```
TomerAberbach: I think mock setup is typically considered part of the arrange step, so should probably move it there for each test

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
     }
 
+    [Fact]
+    public async Task CreateStreamingAggregation_WorksNoContent_RawMessageStartEvent()
+    {
+        // arrange
+        var messagesServiceMock = new Mock<IMessageService>();
+        static async IAsyncEnumerable<RawMessageStreamEvent> GetTestValues()
+        {
+            yield return new(new RawMessageStartEvent(GenerateStartMessage()));
```
TomerAberbach: Just an FYI that we have a codegen change in the pipeline that will add implicit operators from union variant to union, so you will most likely be able to omit many `new(...)` wrappings like this one (and e.g. `Content = new("")` below could be just `Content = ""`)

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
+            yield return new(new RawMessageStopEvent());
+            await Task.CompletedTask;
```
TomerAberbach: What's this for? Is it for suppressing some lint around `async` methods without any `await`?

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
+        }
+        // act
```
TomerAberbach: Typically I see empty newlines between arrange, act, and assert; might be nice to add

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
+        }
+        messagesServiceMock.Setup(e => e.CreateStreaming(It.IsAny<MessageCreateParams>())).Returns(GetTestValues);
+        // act
+        var stream = await messagesServiceMock.Object.CreateStreaming(
+            new()
+            {
+                MaxTokens = 1024,
+                Messages = [new() { Content = new(""), Role = Role.User }],
+                Model = Model.Claude3_7SonnetLatest,
+            }
```
TomerAberbach: Might be nice to extract the params to a static if it's the same for every one of these tests. Makes it easier to tell what matters about the test vs what's just other data

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -38,6 +42,213 @@ public async Task CreateStreaming_Works()
...
+        return new Message()
+        {
+            ID = "Test",
+            Content = [],
+            Model = Model.Claude3OpusLatest,
+            StopReason = StopReason.ToolUse,
+            StopSequence = "",
+            Usage = null
+        };
+    }
```
TomerAberbach: Should this be a static variable instead?

.devcontainer/devcontainer.json
TomerAberbach: Thanks, gonna port this back to codegen :)

src/Anthropic.Client.Tests/Anthropic.Client.Tests.csproj
TomerAberbach, line 1: Seems there a whitespace changes here; can we revert those? That's going to cause lots of merge conflicts with codegen

.gitignore
```diff
@@ -1,2 +1,3 @@
 bin/
 obj/
+.prism.log
```
TomerAberbach: Good catch. Looks like we have this in other SDKs. I'll port this to codegen

src/Anthropic.Client/Services/Beta/BetaMessageContentAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Linq;
+using Anthropic.Client.Models.Beta.Messages;
+using Anthropic.Client.Models.Beta.Messages.BetaCitationsDeltaProperties;
+
+namespace Anthropic.Client.Services.Messages;
+
+/// <summary>
+/// The aggregation model for a stream of <see cref="BetaRawContentBlockDeltaEvent"/>
+/// </summary>
+public class BetaMessageContentAggregator : SseAggregator<BetaRawMessageStreamEvent, BetaMessageAggregationResult>
```
TomerAberbach: Should this be `sealed`?

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,109 @@
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Anthropic.Client;
+
+/// <summary>
+/// Defines the base for all Aggregators using ServerSideStreaming events.
+/// </summary>
+/// <typeparam name="TMessage"></typeparam>
```
TomerAberbach: Is there a point to defining `typeparam` in the documentation if you're not gonna document them? This seems to just be duplicative with what's already in the code

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,109 @@
...
+                {
+                    messages.Add(item);
+                }
+                else if (filterResult == FilterResult.EndMessage)
+                {
+                    break;
+                }
+            }
+
+            if(messages is {Count: 0})
```
TomerAberbach: Do we need to run `./scripts/format`? I would expect this to be `if (messages`

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// </summary>
+    public enum FilterResult
+    {
+        /// <summary>
+        /// The filtered message should be ignored.
+        /// </summary>
+        Ignore = 0,
+
+        /// <summary>
+        /// The message defines the start of a package.
```
TomerAberbach, line 123: What do you mean by "package"?

src/Anthropic.Client/SseAggregator.cs
```diff
@@ -0,0 +1,109 @@
...
+    /// <summary>
+    /// Initialises a new instance of the <see cref="SseAggregator{TMessage, TResult}"/>.
+    /// </summary>
+    /// <param name="messages">The service attached enumerable for reading messages.</param>
+    public SseAggregator(IAsyncEnumerable<TMessage> messages)
+    {
+        this._messages = messages;
+    }
+
+    private Task<TResult>? _collectionTask;
```
TomerAberbach: What's the reason to have this field? It doesn't seem to be read as far as I can tell

src/Anthropic.Client/SseAggregator.cs
```diff
@@ -0,0 +1,109 @@
...
+using System.Threading.Tasks;
+
+namespace Anthropic.Client;
+
+/// <summary>
+/// Defines the base for all Aggregators using ServerSideStreaming events.
+/// </summary>
+/// <typeparam name="TMessage"></typeparam>
+/// <typeparam name="TResult"></typeparam>
+public abstract class SseAggregator<TMessage, TResult>
```
TomerAberbach: Should we make this and its subclasses `internal`? Since people should use it via the extensions anyway?

src/Anthropic/SseAggregator.cs
TomerAberbach, line 1: Not sure where to leave this comment, but do we need any handling or logic related to not allowing these aggregators to be used more than once for the same async enumerable?

src/Anthropic.Client/Services/Messages/MessageAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+using System.Linq;
+using Anthropic.Client.Models.Messages;
+using Anthropic.Client.Models.Messages.CitationsDeltaProperties;
+
+namespace Anthropic.Client.Services.Messages;
+
+/// <summary>
+/// An implementaion of the <see cref="SseAggregator{TMessage, TResult}"/> for aggregating BlockDeltaEvents from the <see cref="IMessageService.CreateStreaming(MessageCreateParams)"/> method.
+/// </summary>
+public class MessageContentAggregator : SseAggregator<RawMessageStreamEvent, MessageAggregationResult>
```
TomerAberbach: Should this be `sealed`?

src/Anthropic.Client/Services/Messages/MessageAggregator.cs
```diff
@@ -0,0 +1,79 @@
...
+        RawContentBlockDeltaEvent _ => FilterResult.Content,
+        RawMessageStopEvent _ => FilterResult.EndMessage,
+        _ => FilterResult.Ignore
+    };
+}
+
+/// <summary>
+/// The aggregation model for a stream of <see cref="RawContentBlockDeltaEvent"/>
+/// </summary>
+public class MessageAggregationResult
```
TomerAberbach: Maybe there was a miscommunication, but the returned aggregated thing should be the same object as the non-streaming method

`Anthropic.Client.Models.Messages.Message` in this case. The idea is that a user may have already written logic that works with the non-streaming endpoint's output. Like imagine they wrote a function that accepts `Message`. It's annoying if they now have to unnecessarily handle this other type instead as a result of switching to streaming.

See how the Java SDK handles this (its aggregation returns the non-streaming method's data model)

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
JPVenson: yesn't its not just for linting but its for when you have an async method that is not async it still needs to return a valid task. So that it runs synchronously

src/Anthropic/SseAggregator.cs
JPVenson: i am trying to think of a scenario where you could do this actually but its hard to come up with. They are "hidden" behind that extension method and you really need to invest some work to do that the way they currently work and what the purpose of the IAsyncEnumerable in general is. 

I dont think thats a big concern but also thats handled via the collectionTask so even if you try to enumerate multiple times, you get the same result and the same operation.

src/Anthropic.Client/SseAggregator.cs
JPVenson: forgot the null coalescing assignment operator good catch

src/Anthropic.Client/SseAggregator.cs
JPVenson: we _can_ but we dont expose any internal that are dangerous to fiddle with and i would prefer at least giving devs the option to implement an aggregator on their own if they want to which could come in handy for some usecases.

src/Anthropic/Helpers/BetaMessageContentAggregator.cs
```diff
@@ -0,0 +1,164 @@
+using System;
```
sd-st, line 1: I think this file belongs somewhere else. Could maybe put it in a new `Helpers` directory under `Anthropic` (btw, we renamed `Anthropic.Client` to `Anthropic` as well)

src/Anthropic.Client/Services/Beta/BetaMessageContentAggregator.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using Anthropic.Models.Beta.Messages;
+
+namespace Anthropic.Services.Messages;
+
+/// <summary>
+/// The aggregation model for a stream of <see cref="BetaRawContentBlockDeltaEvent"/>
+/// </summary>
+public class BetaMessageContentAggregator : SseAggregator<BetaRawMessageStreamEvent, BetaMessage>
```
sd-st: @TomerAberbach I noticed in Java we call this a message accumulator rather than aggregator. I am pretty sure internally we only call it an aggregator; does this make sense or do we want to match Java?

.gitignore
sd-st: Now that this is in codegen we should be able to remove it from this PR

src/Anthropic/SseAggregator.cs
sd-st: I am curious about this also

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,114 @@
...
+    public SseAggregator(IAsyncEnumerable<TMessage> messages)
+    {
+        this._messages = messages;
+    }
+
+    private Task<TResult?>? _collectionTask;
+
+    public virtual Task<TResult?> BeginCollectionAsync()
+    {
+        return _collectionTask ??= Task.Run(async () =>
```
sd-st: Why do we have a separate task here? I also don't understand the point of `_collectionTask` - is this for caching purposes? If so, can't we just cache the accumulated message?

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,114 @@
...
+    /// </summary>
+    /// <param name="messages">The service attached enumerable for reading messages.</param>
+    public SseAggregator(IAsyncEnumerable<TMessage> messages)
+    {
+        this._messages = messages;
+    }
+
+    private Task<TResult?>? _collectionTask;
+
+    public virtual Task<TResult?> BeginCollectionAsync()
```
sd-st: Can we rename this? I think the current name is not super accurate, since it doesn't just being collecting, but follows through to completion

src/Anthropic/SseAggregator.cs
```diff
@@ -0,0 +1,114 @@
...
+            }
+
+            if (messages is { Count: 0 })
+            {
+                return default;
+            }
+
+            if (filterResult != FilterResult.EndMessage)
+            {
+                throw new InvalidOperationException(
```
sd-st: Probably can wrap this in an `AnthropicInvalidDataException` so it's easier to catch

src/Anthropic/Helpers/SseAggregator.cs
```diff
@@ -0,0 +1,114 @@
...
+using System.Threading.Tasks;
+
+namespace Anthropic;
+
+/// <summary>
+/// Defines the base for all Aggregators using ServerSideStreaming events.
+/// </summary>
+/// <typeparam name="TMessage">The raw message base element type.</typeparam>
+/// <typeparam name="TResult">The element type that defines an aggregated <typeparamref name="TMessage"/></typeparam>
+public abstract class SseAggregator<TMessage, TResult>
```
sd-st, line 13: I think we also need this to be an `IAsyncEnumerable` itself? Like one use-case we want to be able to handle, I believe, is that the user may want to lazily use the streaming result, but also accumulate it while they're doing that. I think we'd want something like this:

```c#
var myStream = ...;
var aggregator = new MessageAggregator(myStream);

await foreach (var item in aggregator)
{
   // we may not actually consume the entire stream and/or use the aggregator at all
   if(...) { throw new Exception(); }
}

aggregator.Complete(); // I forget if we decided if we wanted this or not
Console.WriteLine(aggregator.Message);
```

Basically we want to be able to use the stream lazily but still be able to aggregate it at the end. 

@TomerAberbach Am I remembering this correctly? And did we want a manual `Complete` method? I forget what we decided and it's been a while, plus I am having trouble finding it in Slack

src/Anthropic/Helpers/MessageAggregator.cs
```diff
@@ -0,0 +1,130 @@
+using System;
```
sd-st, line 1: FYI we are also going to want to update the README with a code example. You can take a look at Java: https://github.com/anthropics/anthropic-sdk-java

It would also be nice to add an example under the examples folder

sd-st (commented):
Only reviewed some of this PR but just had a few initial comments

src/Anthropic.Client/Services/Beta/BetaMessageContentAggregator.cs
JPVenson: c# has already defined the term "Aggregate" as a linq function that is functionally in the same category so i think it fits better here

src/Anthropic/Helpers/BetaMessageContentAggregator.cs
JPVenson: Yea putting this under helpers might be a good idea.

Also good catch that seemed to be a result of the merge into the new branch.

src/Anthropic/Helpers/SseAggregator.cs
JPVenson: if you want to filter the elements that go into the aggregator, you can still do that with linq 

`GetStreamSource().Where(e => e.Type == "ThatsWhatIWant").Aggregate()`

As the aggregator is but a consumer of __a__ IAsyncEnumerable you can do whatever you want including filtering and modifications to it

src/Anthropic/SseAggregator.cs
JPVenson: this ensures that the provided IAsyncEnumerable is only ever enumerated once as it is a transient provider to the data. This is nessesary as we provide the IAsyncEnumerable via the constructor which could lead to somebody calling this multiple times.

src/Anthropic/SseAggregator.cs
JPVenson: yea i wrote that comment before fully understanding the structure of the streaming protocol^^

src/Anthropic/Helpers/SseAggregator.cs
sd-st: Right but what if you want to also use the streaming output for something, like printing it to the user?

JPVenson: yea funny enough i argued the exact same thing in the notion doc ... 

I think we can still handle that without much change via linq extension methods. Will add that

JPVenson: I refactored the code and made some small changes so it can be more easily added into a linq tree.

```csharp

IAsyncEnumerable<RawMessageStreamEvent> responseUpdates = client.Messages.CreateStreaming(
    parameters
);

// some streaming endpoints have build-in aggregators that create logically aggregated objects that represent the full stream as its counterpart single object.
var message = responseUpdates.Aggregate();

// you can also add an aggregator as part of your linq chain to get realtime streaming and aggregation

var aggregator = new MessageContentAggregator();
await foreach (RawMessageStreamEvent rawEvent in responseUpdates.CollectAsync(aggregator))
{
    // do something with the stream events
    if (rawEvent.TryPickContentBlockDelta(out var delta))
    {
        if (delta.Delta.TryPickThinking(out var thinkingDelta))
        {
            Console.Write(thinkingDelta.Thinking);
        }
        else if (delta.Delta.TryPickText(out var textDelta))
        {
            Console.Write(textDelta.Text);
        }
    }
}

// and then get the full aggregated message
var fullMessage = responseUpdates.Aggregate();

```

scripts/test
JPVenson: i reverted the change as i cannot reproduce the issue anymore

src/Anthropic/Helpers/MessageAggregator.cs
JPVenson: both updated/added

examples/StreamingAggregationExample/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+    ],
+    Model = Model.Claude4Sonnet20250514,
+    Thinking = new ThinkingConfigEnabled() { BudgetTokens = 1024 },
+};
+
+IAsyncEnumerable<RawMessageStreamEvent> responseUpdates = client.Messages.CreateStreaming(
+    parameters
+);
+
+// some streaming endpoints have build-in aggregators that create logically aggregated objects that represent the full stream as its counterpart single object.
```
sd-st:
```suggestion
// some streaming endpoints have built-in aggregators that create logically aggregated objects.
// these represent the full stream as a single object.
```

examples/StreamingAggregationExample/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+var message = await responseUpdates.Aggregate().ConfigureAwait(false);
+
+// you can also add an aggregator as part of your linq chain to get realtime streaming and aggregation
```
sd-st:
```suggestion
// you can also add an aggregator as part of your LINQ chain to get real-time streaming and aggregation
```

examples/StreamingAggregationExample/Program.cs
```diff
@@ -0,0 +1,47 @@
+﻿using Anthropic;
```
sd-st, line 1: Wondering if we should all of the Aggregator stuff to Accumulator so that it matches with Java. @TomerAberbach thoughts?

README.md
```diff
@@ -155,6 +155,46 @@ await foreach (var message in client.Messages.CreateStreaming(parameters))
 }
 ```
 
+### Aggregators
+
+Both the Messages and BetaMessages streaming endpoints have build-in aggregators that can produce the same object as its non-streaming counterparts.
```
sd-st, line 160: I think it would be good to link `Messages` and `BetaMessage` to their respective files, like we do in the Java readme

README.md
```diff
@@ -155,6 +155,46 @@ await foreach (var message in client.Messages.CreateStreaming(parameters))
...
 
+### Aggregators
+
+Both the Messages and BetaMessages streaming endpoints have build-in aggregators that can produce the same object as its non-streaming counterparts.  
+This can be useful when progress should be reported during the generation but then also handle the full result as a single object.
+
+It is possible to either only get the full result object via the `.Aggregate()` extension on the `IAsyncEnumerable` returned by the `CreateStreaming` method or insert an external aggregator into a linq tree:
+
+```csharp
+
```
sd-st: NIT: we can remove this blank line

src/Anthropic/Helpers/SseAggregator.cs
```diff
@@ -0,0 +1,137 @@
...
+{
+    private Dictionary<FilterResult, IList<TMessage>>? _messages;
+    private bool _streamEnded;
+    private TResult? _message;
+
+    /// <summary>
+    /// Collects and filters the provided <see cref="IAsyncEnumerable{TMessage}"/> for aggregation with the <see cref="AggregateAsync"/> method.
+    /// </summary>
+    /// <param name="messageStream">An <see cref="IAsyncEnumerable{TMessage}"/> containing the messages to aggregate.</param>
+    /// <returns></returns>
```
sd-st: Should the `<returns>` be filled in?

src/Anthropic/Helpers/SseAggregator.cs
```diff
@@ -0,0 +1,137 @@
...
+    /// <exception cref="InvalidOperationException">Will be thrown if the aggregator is in an invalid state.</exception>
+    /// <exception cref="AnthropicInvalidDataException">Will be thrown if the aggregator encounters an invalid state of the source message stream.</exception>
+    public virtual async IAsyncEnumerable<TMessage> CollectAsync(
```
sd-st, line 26: What happens if the user only goes through half of this iterator and then calls `AggregateAsync`? I believe we want it to continue through the stream, but it looks like in this case we'll miss the latter half of the streamed response

src/Anthropic/Helpers/MessageAggregator.cs
```diff
@@ -0,0 +1,139 @@
...
+            IEnumerable<T> sources,
+            Func<T, string> expression
+        )
+        {
+            return string.Join(null, (string[])[source, .. sources.Select(expression)]);
+        }
+
+        void As<TDelta>(Func<IEnumerable<TDelta>, ContentBlock> factory)
+        {
+            // those blocks are DELTA variants not the source block
```
sd-st:
```suggestion
            // those blocks are delta variants not the source block
```

Should probably update this everywhere as delta shouldn't be uppercased

src/Anthropic/Helpers/MessageAggregator.cs
```diff
@@ -0,0 +1,139 @@
...
+    {
+        var content = messages[FilterResult.Content].GroupBy(e => e.Index);
+
+        var startMessage =
+            messages[FilterResult.StartMessage]
+                .Select(e => e.Value!)
+                .OfType<RawMessageStartEvent>()
+                .Single()
+            ?? throw new InvalidOperationException(
+                $"Expected to find exactly one {nameof(RawMessageStartEvent)} but found either non or more then one."
```
sd-st:
```suggestion
                $"Expected to find exactly one {nameof(RawMessageStartEvent)} but found either none or more then one."
```

src/Anthropic/Helpers/MessageAggregator.cs
```diff
@@ -0,0 +1,139 @@
...
+    )
+    {
+        var content = messages[FilterResult.Content].GroupBy(e => e.Index);
+
+        var startMessage =
+            messages[FilterResult.StartMessage]
+                .Select(e => e.Value!)
+                .OfType<RawMessageStartEvent>()
+                .Single()
+            ?? throw new InvalidOperationException(
```
sd-st, line 25: `Single` never returns null right? It always throws?

src/Anthropic/Helpers/MessageAggregator.cs
```diff
@@ -0,0 +1,139 @@
...
+                        .. Of<CitationsDelta>()
+                            .Select(e =>
+                                e.Citation.Match<TextCitation>(
+                                    f => f,
+                                    f => f,
+                                    f => f,
+                                    f => f,
+                                    f => f
+                                )
+                            ),
```
sd-st, line 106: Hm, I don't love this but I don't have any better ideas

src/Anthropic/Helpers/BetaMessageContentAggregator.cs
```diff
@@ -0,0 +1,147 @@
+using System;
+using System.Collections.Generic;
```
sd-st, line 2: Note to self that I still need to review this file (have held off for now since it's probably similar to the non-beta file)

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -1,12 +1,46 @@
+using System;
```
sd-st: I think there's a few tests I'd like to see:
- multiple deltas get concatenated together properly
- partially iterating over the aggregated async enumerable and then calling `aggregate` should finish the enumeration and still return the full streamed response

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -1,12 +1,46 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using Anthropic.Models.Beta.Messages;
-using Anthropic.Tests;
+using Anthropic.Models.Messages;
```
sd-st: Hm, I think the non-beta messages stuff belongs in a different file; it doesn't look like we need to mix these together

src/Anthropic/SseAggregatorExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Anthropic;
+
+/// <summary>
+/// Extension methods for providing easy access to Aggregators
+/// </summary>
+public static class SseAggregatorExtensions
+{
+    /// <summary>
+    /// Aggregates all messages received by the streaming event and aggregates them into a single object once the sender indicated a fully delivered stream.
```
sd-st:
```suggestion
    /// Aggregates all messages received by the streaming event and aggregates them into a single object once the sender indicates a fully delivered stream.
```

similar for below

.gitignore
sd-st: looks like it's still here

sd-st (changes requested):

README.md
JPVenson: can you please elaborate?

src/Anthropic/Helpers/SseAggregator.cs
JPVenson: The enumeration pattern in C# hides a bit of boilerplate code. When you actually enumerate an enumerable, you call the `GetEnumerator` method which returns an actual object that is used to go though the elements on a single thread. Multithreadded enumeration is not supported by that pattern so you cannot go though it half way and then call Collect on it.

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
JPVenson: i think the beta tests somehow got lost in a merge and then mixed up umpf

src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
JPVenson: - Multiple deltas are aggregated in `CreateStreamingAggregation_Works`
- If i understand you correctly, that would never be possible. We need to collect the start message in every case without it we are missing certain info that are required by the model. But i think i am just not understanding what you are trying to do

src/Anthropic.Client/Services/Beta/BetaMessageContentAggregator.cs
sd-st: Sure, that makes sense to me

JPVenson:
closed in favor of #105

Issue #4, opened 2025-10-04
Integration with Microsoft.Extensions.AI


KSemenenko:
https://learn.microsoft.com/en-us/dotnet/ai/microsoft-extensions-ai

this will be super usfeull for C# devs, becase this is main abstraction for now for all AI stuff in .NET

KSemenenko:
let me know if you are open for PRs, so I can help with it.

PederHP:
+1 to this. I wouldn't say it is unusable without this support, but it really reduces the scope of what the SDK can be used for. Alternatively you could open a PR in the dotnet repo to add support similar to what is there for some of the other model providers.

In fact, it would be useful to know if your preference is to have the extension support in the SDK or to have it in an extensions package - because the dotnet repo is open-source, so if you prefer a separate package, then the community can begin adding the support already.

KSemenenko:
I did similar, https://github.com/managedcode/together-dotnet/tree/main/Together

So we have nice rich client and this client also can be IChatClient so you can use it in any ways.

sd-st:
Sorry, I meant to follow up earlier. This is something we have planned for the near future, thanks!

jeremy-schaab:
I couldn't wait https://github.com/jeremy-schaab/Microsoft.Extensions.AI.Anthropic

sd-st:
@jeremy-schaab FYI #10 should add this support and is pretty close to being merged, but you could use that branch if you wish

jeremy-schaab:
@sd-st Thanks for the update, I will wait until it's released to switch over.

sd-st:
All done! Thanks @stephentoub

PR #10, opened 2025-11-05
Add Microsoft.Extensions.AI support

stephentoub:
Fixes https://github.com/anthropics/anthropic-sdk-csharp/issues/4

This provides implementations of `IChatClient` for both the main and beta clients. The actual src additions are a few hundred lines; most of the PR is extensive tests covering using the anthropic clients via IChatClient, providing coverage of both.


.gitignore
```diff
@@ -1,2 +1,3 @@
 bin/
 obj/
+.vs
```
sd-st: NIT: Generally I think things like this belong in a personal git excludesfile, otherwise over time the gitignore will get polluted by all the different editors people use. I checked the other Anthropic SDKs and we don't have this in their gitignores either

src/Anthropic.Client/Anthropic.Client.csproj
```diff
@@ -30,6 +30,6 @@
     <None Include="..\..\README.md" Pack="true" PackagePath="\"/>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Text.Json" Version="9.0.9"/>
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="9.10.2" />
```
sd-st: I think we should leave `System.Text.Json` in. My concern is that if our code generation logic changes, or Anthropic's schema changes, we may need the .NET 9 JSON extensions.

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
...
+
+        /// <inheritdoc />
+        void IDisposable.Dispose() { }
+
+        /// <inheritdoc />
+        public object? GetService(Type serviceType, object? serviceKey = null)
+        {
+            ArgumentNullException.ThrowIfNull(serviceType);
+
+            if (serviceKey is null)
```
sd-st: Could we do a `if(serviceKey is null) { return null; }` so that we don't have an extra layer of indentation?

src/Anthropic/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
...
+
+                foreach (AIContent content in message.Contents)
+                {
+                    switch (content)
+                    {
+                        case TextContent tc:
+                            string text = tc.Text;
+                            if (message.Role == ChatRole.Assistant)
+                            {
+                                text = text.TrimEnd();
```
sd-st, line 458: You are the expert of course, but it seems a bit weird to me to be modifying the string here and not just pass it around verbatim. Is this typical?

src/Anthropic/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
...
+                                if (!string.IsNullOrWhiteSpace(text))
```
sd-st, line 459: I kind of feel the same about using `IsNullOrWhitespace` instead of just `is null`. But I'll defer to your judgement

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
...
+        private static UsageDetails ToUsageDetails(BetaMessageDeltaUsage usage) =>
+            ToUsageDetails(usage.InputTokens, usage.OutputTokens, usage.CacheCreationInputTokens, usage.CacheReadInputTokens);
+
+        private static UsageDetails ToUsageDetails(long? inputTokens, long? outputTokens, long? cacheCreationInputTokens, long? cacheReadInputTokens)
+        {
+            UsageDetails usageDetails = new()
+            {
+                InputTokenCount = inputTokens,
+                OutputTokenCount = outputTokens,
+                TotalTokenCount = (inputTokens ?? 0) + (outputTokens ?? 0),
```
sd-st: If they're both null, do we want to make this null too? Maybe too edge-case-y to matter, just wanted to bring it up in case

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
...
+    /// <returns>An <see cref="IChatClient"/> that can be used to converse via the <see cref="IMessageService"/>.</returns>
+    /// <exception cref="ArgumentNullException"><paramref name="messageService"/> is <see langword="null"/>.</exception>
+    public static IChatClient AsIChatClient(this IMessageService messageService, string? defaultModelId = null)
+    {
+        ArgumentNullException.ThrowIfNull(messageService);
+
+        return new AnthropicChatClient(messageService, defaultModelId);
+    }
+
+    private sealed class AnthropicChatClient(IMessageService messageService, string? defaultModelId) : IChatClient
```
sd-st: How would you feel about this taking an `AnthropicClient` instead?

- The fact that we're using the `MessageService` seems like it isn't super relevant to downstream users of this class
- We wouldn't have to make `_client` on `MessageService` internal

src/Anthropic.Client.Tests/AnthropicClientBetaExtensionsTests.cs
```diff
@@ -0,0 +1,278 @@
+using Anthropic.Client;
+using Anthropic.Client.Services.Beta;
+using Anthropic.Client.Services.Beta.Messages;
+using System;
+using System.Threading.Tasks;
+
+#nullable enable
```
sd-st: We have `<Nullable>enable</Nullable>` in the csproj so I believe this is a no-op

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
```diff
@@ -0,0 +1,627 @@
+using Anthropic.Client.Core;
```
sd-st: Does this beta extension have any additional functionality over the non-beta extension? If not I'm wondering if it's better if we don't do this and just have the users always use the non-beta endpoint

sd-st (changes requested):
Hey, thanks for the PR!

I left some comments; I'm not super familiar with `Microsoft.Extensions.AI` so some comments are probably a little silly. Overall though this looks really good, just had a few minor points.

Aside from the comments:
- We should probably add this to the README, and also put an example for it in `examples`
- Some lints are failing (You can check this with `./scripts/lint`)

.gitignore
stephentoub: I can certainly remove it if you think it doesn't belong. Anyone using Visual Studio, though, ends up with tons of artifacts, and as VS is such a common IDE for .NET, I expect you'll find these files ending up in PRs as the project gains traction. I'll remove it, though.

src/Anthropic.Client/Anthropic.Client.csproj
stephentoub: I can put it back. I'd removed it purely because M.E.AI.Abstractions itself references STJ on platforms where it's not in the shared framework. I'm curious, though... is there something from the 9.x version that you're using that's not in the net8.0 version that's part of netcoreapp? The library is currently targeting net8.0 only, which already has STJ as part of it, so the only reason you'd need to reference a newer version is if there was functionality you needed in the newer version.

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
stephentoub: You mean `if (serviceKey is not null) { return null; }`? Sure, I was just coalescing the returns so there was a single return null site, but I can invert it.

src/Anthropic/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
stephentoub: We don't _have_ to do this. The problem is that Anthropic's Claude models are _really_ sensitive to the content that shows up in assistant messages; they don't produce assistant messages that end in spaces, and they then fail if you give them assistant messages that ends in spaces. If you're just roundtripping to Anthropic then, it's fine, and this isn't needed. But if you take the output from one model, where assistant messages may end in spaces, and pass it off to a Claude, it blows up, and that shows up in agentic scenarios where you're dealing with agents backed by models from different providers. If you're concerned, we can remove it.

src/Anthropic.Client.Tests/AnthropicClientBetaExtensionsTests.cs
stephentoub: It's in the src project, but I don't see it in the tests project. Am I missing it?

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
stephentoub: In particular there are tools in the beta endpoint that don't exist in the non-beta endpoint, e.g. code interpreter and MCP. I tried to find a good way to have the beta IChatClient just augment the non-beta one with the bit of additional functionality, but the two underlying object models in AnthropicClient are entirely different / duplicated, such that all of the types involved have different identities... effectively two completely different clients. Have you considered deduplicating those? It would make it a lot easier to switch between the endpoints, experiment with new beta features without needing to entirely rewrite your client code, and would enable the two IChatClients to be structured without all this duplication.

That said, again, it's your project, so if you don't think it's worth it, we can just get rid of the beta one.

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
stephentoub: > How would you feel about this taking an AnthropicClient instead?

I'm not sure how that would work... there's already an AsIChatClient that takes an AnthropicClient for targeting the non-beta endpoint. Are you suggesting this AsIChatClient accept an AnthropicClient but be named differently, e.g. AsBetaIChatClient or something like that? As it stands in the PR, if you call AsIChatClient on an anthropic client, you'll get this:
https://github.com/anthropics/anthropic-sdk-csharp/pull/10/files/3bebd2b459df5db1bac963e9e99715ef6841ecde#diff-eb1e594d1472c137248ab12fcee6ec12c07e3317f9390d953efa1c2862d5377bR24

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
stephentoub: Sure, we can do that.

stephentoub:
@sd-st, I just merged in the latest in main, and there are a few changes that are quite cumbersome. In particular, making all of the properties on MessageCreateParams be `init` is really hard to work with. Take a look at the GetMessageCreateParams in this PR and consider the changes that I'm now having to make in order for this to work. I'm going to do so, but please reconsider having these be init-only.

stephentoub:
Actually, @sd-st, when you combine the init-only aspect with the fact that all of the init's always set a value for the specified property even when storing null, it's basically impossible to now do the right thing, as you can't conditionally set an init.

stephentoub:
Looks like I can work around it for now by just using the raw properties and bypassing the strongly-typed API altogether...

src/Anthropic.Client/Anthropic.Client.csproj
stephentoub: I see now that the tests are using new 9.x surface area.

TomerAberbach:
I'll let Stephen take a look tomorrow and give his opinion as well, but I think:

- In most cases the user can compute the necessary info/data before/in subroutines and pass those results when constructing the immutable data model
- In some cases it may be appropriate to use [`with` expressions](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/with-expression) to create new instances
- In a limited number of ultra dynamic cases, using the raw properties may be appropriate

I've not looked deeply at this PR so not sure which one applies, but just wanted to note that

stephentoub:
> In most cases the user can compute the necessary info/data before/in subroutines and pass those results when constructing the immutable data model

They can, and when it's just `init` in the mix, you can often make it work. But it's compounded by how these implementations currently work. Consider for example MessageCreateParams.Temperature:
```csharp
    public double? Temperature
    {
        get
        {
            if (!this._bodyProperties.TryGetValue("temperature", out JsonElement element))
                return null;

            return JsonSerializer.Deserialize<double?>(element, ModelBase.SerializerOptions);
        }
        init
        {
            this._bodyProperties["temperature"] = JsonSerializer.SerializeToElement(
                value,
                ModelBase.SerializerOptions
            );
        }
    }
```
It's nullable, but if you set `Temperature = null;`, that's still going to write a `JsonElement` into `this._bodyProperties["temperature"]`. Because the property is `init`, if I want to write a routine like:
```csharp
public static MessageCreateParams Create(double? temperature)
{
   ...
}
```
I can't implement it like this because of the `init`:
```csharp
public static MessageCreateParams Create(double? temperature)
{
    MessageCreateParams p = new();
    if (temperature is not null)
    {
        p.Temperature = temperature;
    }
    return p;
}
```
and I can't write it like this because doing so populates the MessageCreateParams with an empty temperature, which will result in `"temperature":null` being emitted in the JSON when it's serialized:
```csharp
public static MessageCreateParams Create(double? temperature)
{
    return new()
    {
        Temperature = temperature,
    };
}
```
which means my option is to write it like:
```csharp
public static MessageCreateParams Create(double? temperature)
{
    return temperature is not null ?
        new()
        {
            Temperature = temperature,
        } :
        new();
}
```
That's fine if I just have the one property to set. But if I have two, now I have four possible ways I need to construct the instance. Three properties, eight. Etc. It's exponential.

I could, as you suggest, with these now being records, use `with`, but that ends up creating a whole new object and subdictionaries and so on for every property that gets set (since each would be done one at a time). Maybe the overhead wouldn't be prohibitive, especially given all the other expenses involved currently with roundtripping to JSON for each get/set, but it's an unfortunate expense, and O(N^2) allocation in terms of the number of properties set (given the dictionaries involved).

I pushed a commit that drops down to just working with the raw JSON. But from a usability perspective, it's quite poor.

Separate from this PR and just for general usability of the library, I'd urge reconsidering `init` and/or making the codegen more amenable to `null`s being stored. If I could store `null` and have it be a nop when already null, then I could just have a shadow local for every property and construct the instance once with all of its properties set to those locals. I can't do that now, though, due to the aforementioned issue of setting `null` not being a nop.

TomerAberbach:
Ah, hmmm, for properties that can only be omitted or set to non-null, I think we should turn `null` init into omitted. For properties that can be both null or omitted, we can't do that

I'll need to double check the codegen implementation tomorrow to see if that's what we're currently doing (can't tell from this PR cause I'm not sure off the top of my head what the schema of each property is). If it's not what we're doing, then I'll fix that, which should help in most cases (it's rare-ish for an API to have a property that can be both null or omitted)

.gitignore
felixfbecker: @sd-st this seems fine to me, we should make it easy .NET contributors if this is the most common IDE for them

sd-st: Sure, I don't feel that strongly

sd-st: Actually, I'm going to backport this to codegen. Just an FYI that you'll get a (very minor) merge conflict as a result

src/Anthropic.Client/Anthropic.Client.csproj
sd-st: Yeah I think the main thing we need it for is deep comparisons of JsonElements

src/Anthropic/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
sd-st: I see, that makes sense. I'm happy for it to be left in then!

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
sd-st: I see. Generally I think users use the `AnthropicClient` directly so it should be pretty rare for a user to have access to a `MessageService` but not an underlying `AnthropicClient`. That said, we can still leave this in

.gitignore
stephentoub: Sounds good, thanks

src/Anthropic.Client.Tests/AnthropicClientBetaExtensionsTests.cs
sd-st: You're totally right. I'm going to backport that to codegen as well

src/Anthropic.Client/Services/Beta/Messages/AnthropicClientBetaExtensions.cs
sd-st: Totally makes sense, we should keep it then!

sd-st:
@stephentoub FYI many of the above improvements we discussed (including init for non-required properties) have been put into `next`, so if you rebase off of that branch you should be good!

stephentoub:
Thanks, @sd-st. Updated and rebased on next.

src/Anthropic/Anthropic.csproj
```diff
@@ -30,6 +30,7 @@
     <None Include="..\..\README.md" Pack="true" PackagePath="\"/>
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Text.Json" Version="9.0.9"/>
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="10.0.0" />
+    <PackageReference Include="System.Text.Json" Version="10.0.0"/>
```
sd-st: I would probably keep the Json import identical (including version number) so we dont run into merge conflicts down the line

src/Anthropic.Client/AnthropicClientExtensions.cs
```diff
@@ -0,0 +1,666 @@
...
+        public object? GetService(System.Type serviceType, object? serviceKey = null)
+        {
+            ArgumentNullException.ThrowIfNull(serviceType);
+
+            return
+                serviceKey is not null ? null :
+                serviceType == typeof(ChatClientMetadata) ? _metadata ??= new ChatClientMetadata("anthropic", _anthropicClient.BaseUrl, _defaultModelId) :
+                serviceType.IsInstanceOfType(_anthropicClient) ? _anthropicClient :
+                serviceType.IsInstanceOfType(this) ? this :
+                null;
```
sd-st: NIT: This is quite a bit of logic to squeeze into a single statement. I think it would be good to break it out a bit so that it's easier to understand

src/Anthropic.Client/AnthropicClientExtensions.cs
```diff
@@ -0,0 +1,666 @@
...
+
+                if (contents.Count == 0)
+                {
+                    continue;
+                }
+
+                messageParams.Add(new()
+                {
+                    Role = message.Role == ChatRole.Assistant ? Role.Assistant : Role.User,
+                    Content = new(contents),
```
sd-st: We no longer need this `new` wrapper - we have implicit operators that'll take care of this now FYI

src/Anthropic.Client/AnthropicClientExtensions.cs
```diff
@@ -0,0 +1,666 @@
...
+            {
+                messageParams.Add(new() { Role = Role.User, Content = new("\u200b") }); // zero-width space
+            }
+
+            return messageParams;
+        }
+
+        private MessageCreateParams GetMessageCreateParams(List<MessageParam> messages, List<TextBlockParam>? systemMessages, ChatOptions? options)
+        {
+            Dictionary<string, JsonElement>? bodyProperties = null;
```
sd-st: Hmm generally I think we should prefer not to deal with bodyProperties directly except when trying to mess with properties in a way that doesn't conform to the schema (which should be quite rare)

Wondering if we can do something else here. I personally don't really have an issue with conditionally chaining `with` statements - there will be some unnecessary copies involved but they should be quite small. Open to other ideas as well

sd-st (dismissed):
Aside from the comments, there are some failing lints (`./scripts/lint`) and it probably makes sense to add something about this to the readme and examples

src/Anthropic/Anthropic.csproj
stephentoub: Microsoft.Extensions.AI.Abstractions references STJ 10, which leads to nuget restore errors if the project tries to downgrade.

sd-st: I see, in that case can leave it as-is

stephentoub:
Thanks, @sd-st. All feedback addressed and merged with latest next.

stephentoub:
Ah, I missed your comment about linting errors... where can I see them?

sd-st:
> Ah, I missed your comment about linting errors... where can I see them?

If you run `./scripts/lint` you should be able to see them

stephentoub:
Ok, done. Thanks.

stephentoub:
@sd-st, I've rebased back on main now that it's been updated with everything from next.

Anything else I can do to help push this forward?

Thanks!

src/Anthropic/Services/Beta/MessageService.cs
```diff
@@ -80,6 +80,11 @@ public async IAsyncEnumerable<BetaRawMessageStreamEvent> CreateStreaming(
         var bodyProperties = parameters.BodyProperties.ToDictionary(e => e.Key, e => e.Value);
         bodyProperties["stream"] = JsonSerializer.Deserialize<JsonElement>("true");
 #endif
+        parameters = MessageCreateParams.FromRawUnchecked(
+            parameters.HeaderProperties,
+            parameters.QueryProperties,
+            bodyProperties
```
stephentoub, line 86: @sd-st, the lack of this was caught by my tests. I'm assuming this is supposed to be here?

sd-st:
Hey @stephentoub - sorry about the long delay here. Haven't forgotten about this, just have some fairly high pri stuff on my plate. I'm hoping to get to this over the next few days but I'm off on Friday so I can't make any promises.

stephentoub:
Sounds good. Thanks.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -0,0 +1,801 @@
...
+                        Output = mcpToolResult.Content.Value switch
+                        {
+                            string s => [new TextContent(s)],
+                            IReadOnlyList<BetaTextBlock> texts => texts.Select(FromBetaTextBlock).ToList(),
+                            _ => [],
+                        },
+                        RawRepresentation = mcpToolResult,
+                    };
+
+                case BetaCodeExecutionToolResultBlock ce when ce.Content?.Value is BetaCodeExecutionResultBlock ceOutput:
```
rogerbarreto: Tool Result Error is not being captured here and I would make a distinction between what is an `ToolError` vs `CodeError`, not sure if both should be represented as `ErrorContent`, maybe only tool, not the code execution.

```suggestion
case BetaContentBlock ct when ct.TryPickCodeExecutionToolResult(out var cer):
{
    var codeResult = new CodeInterpreterToolResultContent() { Outputs = [] };
    if (cer.Content.TryPickError(out var cerErr))
    {
        codeResult.Outputs.Add(new ErrorContent(null) { ErrorCode = cerErr.ErrorCode.Value().ToString() });
    }
    if (cer.Content.TryPickResultBlock(out var cerResult))
    {
        codeResult.Outputs.Add(new TextContent(cerResult.Stdout) { RawRepresentation = cerResult });
        if (!string.IsNullOrWhiteSpace(cerResult.Stderr))
        {
            codeResult.Outputs.Add(new TextContent(cerResult.Stderr) { RawRepresentation = cerResult });
        }

        if (cerResult.Content is { Count: > 0 })
        {
            foreach (var ceOutputContent in cerResult.Content)
            {
                codeResult.Outputs.Add(new HostedFileContent(ceOutputContent.FileID));
            }
        }
    }

    contents.Add(codeResult);
    break;
}
```

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -0,0 +1,801 @@
...
+                    return new McpServerToolCallContent(mcpToolUse.ID, mcpToolUse.Name, mcpToolUse.ServerName)
+                    {
+                        Arguments = mcpToolUse.Input.ToDictionary(e => e.Key, e => (object?)e.Value),
+                        RawRepresentation = mcpToolUse,
+                    };
+
+                case BetaMCPToolResultBlock mcpToolResult:
+                    return new McpServerToolResultContent(mcpToolResult.ToolUseID)
+                    {
+                        Output = mcpToolResult.Content.Value switch
```
rogerbarreto: We might want to do tool error checking.

```suggestion
                        Output = mcpToolResult.IsError 
                            ? [new ErrorContent(mcpToolResult.Content.Value.ToString())]
                            : mcpToolResult.Content.Value switch
```

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -0,0 +1,801 @@
...
+            CitationAnnotation annotation = new()
+            {
+                Title = citation.Title,
+                Snippet = citation.CitedText,
+                FileId = citation.FileID,
+                AnnotatedRegions = [new TextSpanAnnotatedRegion()
+                {
+                    StartIndex = (int?)citation.StartBlockIndex,
+                    EndIndex = (int?)citation.EndBlockIndex,
+                }]
```
rogerbarreto: Not all the citations are `TextSpanAnnotatedRegion` they have other representations that we might erronealy represent as the `TextSpan`.


```suggestion
case BetaTextCitation bChar when bChar.TryPickCitationCharLocation(out var charLocation):
{
    annotation.AnnotatedRegions.Add(new TextSpanAnnotatedRegion { StartIndex = (int?)charLocation?.StartCharIndex, EndIndex = (int?)charLocation?.EndCharIndex });
    break;
}

case BetaTextCitation search when search.TryPickCitationSearchResultLocation(out var searchLocation) && Uri.IsWellFormedUriString(searchLocation.Source, UriKind.RelativeOrAbsolute):
{
    annotation.Url = new Uri(searchLocation.Source);
    break;
}

case BetaTextCitation search when search.TryPickCitationsWebSearchResultLocation(out var searchLocation):
{
    annotation.Url = new Uri(searchLocation.URL);
    break;
}
```

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
rogerbarreto: We might use the same implementation for CodeExecution with the `BashCodeExecution` tool as well.

```suggestion
                case BetaContentBlock ct when ct.TryPickBashCodeExecutionToolResult(out var bashCer):
                case BetaCodeExecutionToolResultBlock ce when ce.Content?.Value is BetaCodeExecutionResultBlock ceOutput:
```

sd-st:
FYI, we will probably still want to merge this into `next` I think

src/Anthropic/Services/Beta/MessageService.cs
sd-st: Yes, thank you! Looks like it got cut out accidentally during a merge conflict

src/Anthropic.Client/Anthropic.Client.csproj
```diff
@@ -0,0 +1,36 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
sd-st: I suspect this probably shouldn't be here

aside: sorry about how much we've completely changed the structure of this repository under your feet 😬

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -0,0 +1,722 @@
...
+        this IAnthropicClient client,
+        string? defaultModelId = null,
+        int? defaultMaxOutputTokens = null)
+    {
+        if (client is null)
+        {
+            throw new ArgumentNullException(nameof(client));
+        }
+
+        if (defaultMaxOutputTokens is <= 0)
```
sd-st, line 43: Just curious, why the `is` here? Seems like `null <= 0` returns false but `null is <= 0` blows up - don't we want the former behaviour?

stephentoub:
> FYI, we will probably still want to merge this into `next` I think

Ok :) Rebased back on next.

src/.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+# C# files
+[*.cs]
+
+# Disabled due to multitargeting
+dotnet_diagnostic.CA1510.severity = none      # CA1510: Use ArgumentNullException throw helper
```
sd-st: FYI we also have a .editconfig in the root you could put this in

sd-st (commented):
Looks pretty good! Just had a few minor comments. As an aside, the formatting looks a tiny bit weird in some places - I suspect a `./scripts/format` should clean things up

src/Anthropic/AnthropicClientExtensions.cs
stephentoub: Either is fine, neither will blow up, it's really just stylistic:

https://lab.razor.fyi/#fY-xagJBEIbxSGGmCj7BlKZZYtpTgyiESIRArKycW9fL4N5ucrMnHkHIO9pLXiDPYIqzCEfO7p9_vq_44ScCeMl9mlOmtHSOUSHsUnwtJZgshr-XGntrjQ7snahH40zO-gLxlGVFoMSaGjNhSp2XwFqaP2rsV2bkyJbCdeyZ3Uetmr_lhlbs0qZezUk2EgNoSyJ4XgyfgIj4XiSWNUqgwBoT7y3Oel124QG3ZAtzi4NhlZAF-wO8ixu9-_-9StrDNNKyuG5_HU7fh_XN1bK1a_0C

I can change it if you'd like.

src/Anthropic.Client/Anthropic.Client.csproj
stephentoub: No worries.

src/Anthropic/AnthropicClientExtensions.cs
sd-st: Sorry, you are right. I was testing in a REPL in a naive way. I am good to leave this as-is

sd-st: Also TIL about lab.razor.fyi - very cool

sd-st:
@stephentoub I got a pretty big diff from `./scripts/format`. It looks like a lot of that is unrelated to this PR, but if you could autoformat the changed files with `dotnet csharpier format path/to/file` that would be appreciated. LGTM other than that!

stephentoub:
> @stephentoub I got a pretty big diff from `./scripts/format`. It looks like a lot of that is unrelated to this PR, but if you could autoformat the changed files with `dotnet csharpier format path/to/file` that would be appreciated. LGTM other than that!

Done, thanks.

sd-st:
@stephentoub Hey, really sorry about all the merge conflicts. If you want to do one more conflict fix + formatting pass (if it needs it) I'll be standing by to give it a quick review + merge. Excited to get this in (:

sd-st (approved):
Yay! Thank you for this PR (:

stephentoub:
Thank you, @sd-st.

PR #18, opened 2025-11-19
fix(client): return correct type for foundry#WithOptions

### LINKED ISSUE #17: Foundry Client Issue
This code is creating a AnthropicClient when it should be creating a AnthropicFoundryClient. The with options creates an AnthropicClient and then the BeforeSend in the AnthropicFoundryClient is not called. Removing             ._client.WithOptions(options =>
                options with
                {
                    Timeout = options.Timeout ?? TimeSpan.FromMinutes(10),
                }
            ) fixes the issue but still problematic.

```
    public async Task<Message> Create(
        MessageCreateParams parameters,
        CancellationToken cancellationToken = default
    )
    {
        HttpRequest<MessageCreateParams> request = new()
        {
            Method = HttpMethod.Post,
            Params = parameters,
        };
        using var response = await this
            ._client.WithOptions(options =>
                options with
                {
                    Timeout = options.Timeout ?? TimeSpan.FromMinutes(10),
                }
            )
            .Execute(request, cancellationToken)
            .ConfigureAwait(false);
        var message = await response.Deserialize<Message>(cancellationToken).ConfigureAwait(false);
        if (this._client.ResponseValidation)
        {
            message.Validate();
        }
        return message;
    }
```

JPVenson:
fixes #17

### DIFF
--- a/src/Anthropic/AnthropicClient.cs
+++ b/src/Anthropic/AnthropicClient.cs
@@ -14,6 +14,21 @@ namespace Anthropic;
 
 public class AnthropicClient : IAnthropicClient
 {
+    public AnthropicClient()
+    {
+        _options = new();
+
+        _messages = new(() => new MessageService(this));
+        _models = new(() => new ModelService(this));
+        _beta = new(() => new BetaService(this));
+    }
+
+    public AnthropicClient(ClientOptions options)
+        : this()
+    {
+        _options = options;
+    }
+
     static readonly ThreadLocal<Random> _threadLocalRandom = new(() => new Random());
 
     static Random Random
@@ -21,7 +36,7 @@ public class AnthropicClient : IAnthropicClient
         get { return _threadLocalRandom.Value!; }
     }
 
-    readonly ClientOptions _options;
+    protected readonly ClientOptions _options;
 
     public HttpClient HttpClient
     {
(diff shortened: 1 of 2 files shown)


packyg (dismissed):

src/Anthropic/AnthropicClient.cs
```diff
@@ -14,14 +14,29 @@ namespace Anthropic;
...
+        _messages = new(() => new MessageService(this));
+        _models = new(() => new ModelService(this));
+        _beta = new(() => new BetaService(this));
+    }
+
+    public AnthropicClient(ClientOptions options)
+        : this()
+    {
+        _options = options;
+    }
```
TomerAberbach: Let's not move this. It's going to cause merge conflicts with codegen whenever we change the constructors in codegen

TomerAberbach: If you think we should be moving the constructors elsewhere in the ordering for style reasons then let me know and I'll move them in codegen

src/Anthropic/AnthropicClient.cs
```diff
@@ -301,7 +301,7 @@ static bool ShouldRetry(Exception e)
     {
         return e is IOException || e is AnthropicIOException;
     }
-
+
```
TomerAberbach: nit: revert this as well

TomerAberbach: Looks like you done it :)

TomerAberbach (approved):

src/Anthropic/AnthropicClient.cs
JPVenson: yes constructors generally come after private and static members but before all other methods and properties. See AnthropicFoundryClient.cs:22

TomerAberbach: Cool, I'll file a ticket to do that, thanks :)

Issue #20, opened 2025-11-20
.Net Standard 2.0 Compatibility Issue


rogerbarreto:
The current Shim Technique is creating an issue while attempting to use the project with other solutions that the namespace / types collide with the Shimmed types in [GlobalShims.cs](https://github.com/anthropics/anthropic-sdk-csharp/blob/main/src/Anthropic/GlobalShims.cs)

i.e:
```powershell
AnthropicChatCompletion.IntegrationTests net472 failed with 1 error(s) (0.6s)
    CSC : error CS1685: The predefined type 'ImmutableArray<T>' is defined in multiple assemblies in the global alias;
 using definition from 'Anthropic, Version=10.1.2.0, Culture=neutral, PublicKeyToken=null'
```

Suggestion:

1. Add `.NetStandard` conditional dependencies csproj for the `System.Collections.Immutable` libraries
2. Update the `Shims` to be `internal` only, so no collision happen in the caller side.

PR #21, opened 2025-11-20
Fix incorrect handling of polyfills

stephentoub:
- The System.Runtime.CompilerServices / System.Diagnostics.CodeAnalysis types should all be internal rather than public; otherwise they will lead to ambiguity warnings with consumers and also won't type unify.
- The System.Collection.Immutable / ValueTask types should not be polyfilled; they instead should just use the netstandard2.0 packages.
- Anthropic.Foundry needs to include the shims.cs file correctly.

cc: @sd-st, @rogerbarreto

### DIFF
--- a/src/Anthropic.Foundry/Anthropic.Foundry.csproj
+++ b/src/Anthropic.Foundry/Anthropic.Foundry.csproj
@@ -13,6 +13,10 @@
     <PackageReference Include="Microsoft.Rest.ClientRuntime.Azure.Authentication" Version="2.4.1" />
     <PackageReference Include="Azure.Identity" Version="1.17.0" />
   </ItemGroup>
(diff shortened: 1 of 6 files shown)


rogerbarreto:
@stephentoub Thanks! That's great!

- Fixes #20

rogerbarreto (approved):

sd-st:
This could use a rebase if you don't mind (:

stephentoub:
Done

sd-st (approved):

sd-st:
Sorry I realized we were targeting the wrong branch 😅

stephentoub:
Rebased on next.

What is the purpose of main and next then?

All changes should target next and then main is effectively a release branch?

sd-st:
Yup exactly. We only merge into main when we cut a new release

sd-st (approved):

PR #22, opened 2025-11-20
refactor(client): Add csharpier check to lint script

JPVenson:
Adds csharpier call to the lint script


sd-st:
FYI I have a PR up to do this as part of codegen, which should be merged shortly

sd-st:
Closing as this is in codegen now

Issue #27, opened 2025-11-21
KeyNotFound errors while using in `netstandard2.0` build.


rogerbarreto:
Attempting to use the services in `Net Standard` environments is throwing exceptions in the given line.

<img width="1031" height="417" alt="Image" src="https://github.com/user-attachments/assets/224aeb84-f2ad-440a-a67d-dee13e9751ed" />

Suggest adding `net462` or `net472` targets for the Test project to capture such errors.

Stack:

```
 Stack Trace: 
ThrowHelper.ThrowKeyNotFoundException()
Dictionary`2.get_Item(TKey key)
ModelConverter`1.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
JsonConverter`1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state)
JsonTypeInfo`1.Deserialize(Utf8JsonReader& reader, ReadStack& state)
JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 utf8Json, JsonTypeInfo`1 jsonTypeInfo, Nullable`1 actualByteCount)
JsonSerializer.Deserialize[TValue](JsonElement element, JsonSerializerOptions options)
BetaRawMessageStreamEventConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options)
JsonConverter`1.TryRead(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options, ReadStack& state, T& value, Boolean& isPopulatedValue)
```

JPVenson:
Hey, does that happen for all methods/endpoints or only particular ones?

PR #29, opened 2025-11-21
Enable netstandard test coverage

rogerbarreto:
This pull request introduces multi-framework testing support for the project, allowing tests to run against both `.NET 8.0` and `.NET Framework 4.6.2`. The changes update the test scripts, CI workflow, and test project configuration to enable this functionality, and include minor code cleanups in the test base class.

**Multi-framework testing support:**

* Updated `src/Anthropic.Tests/Anthropic.Tests.csproj` to target both `net8.0` and `net462` frameworks, and set the language version to 12.
* Enhanced the `scripts/test` script to accept a `--framework` argument, defaulting to `net8.0`, and passing the specified framework to `dotnet test`. [[1]](diffhunk://#diff-d7ad7c7666dae3e86d4d3a850f7c4df6dcde86454e9995001ea017e44a3d35d5R12-R28) [[2]](diffhunk://#diff-d7ad7c7666dae3e86d4d3a850f7c4df6dcde86454e9995001ea017e44a3d35d5L56-R73)

**Continuous Integration workflow updates:**

* Modified `.github/workflows/ci.yml` to run tests for both `net8.0` and `net462` using the updated test script and added a new `test-netstandard` job for `net462`.

**Test code cleanup:**

* Fixed a namespace import in `AnthropicClientExtensionsTestsBase.cs` and removed a redundant import. [[1]](diffhunk://#diff-dcc8ae7a379bf0e30728ddfac2154a44dd543e968d60acbc45fddcdea650e67cR1) [[2]](diffhunk://#diff-dcc8ae7a379bf0e30728ddfac2154a44dd543e968d60acbc45fddcdea650e67cL10)
* Removed an unused `CancellationToken` parameter from a `ReadAsStringAsync` call in the test base class.

### DIFF
--- a/src/Anthropic.Tests/Anthropic.Tests.csproj
+++ b/src/Anthropic.Tests/Anthropic.Tests.csproj
@@ -1,11 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net462</TargetFrameworks>
+    <TargetFrameworks>net9.0;net462</TargetFrameworks>
     <Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
     <ImplicitUsings>disable</ImplicitUsings>
-    <LangVersion>12</LangVersion>
+    <LangVersion>13</LangVersion>
     <!-- Disable skipped tests warnings -->
     <NoWarn>$(NoWarn),xUnit1004</NoWarn>
     <!-- Disable deprecation warnings since we may reference our own deprecated symbols -->
--- a/src/Anthropic.Tests/Anthropic.Tests.csproj
+++ b/src/Anthropic.Tests/Anthropic.Tests.csproj
@@ -1,10 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
     <ImplicitUsings>disable</ImplicitUsings>
+    <LangVersion>12</LangVersion>
     <!-- Disable skipped tests warnings -->
     <NoWarn>$(NoWarn),xUnit1004</NoWarn>
     <!-- Disable deprecation warnings since we may reference our own deprecated symbols -->
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -1,3 +1,4 @@
+using Anthropic;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -7,7 +8,6 @@ using System.Text.Json;
 using System.Text.Json.Nodes;
 using System.Threading;
 using System.Threading.Tasks;
-using Anthropic;
 
 #pragma warning disable IDE0130 // Namespace does not match folder structure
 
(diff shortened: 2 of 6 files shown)


src/Anthropic.Tests/Anthropic.Tests.csproj
```diff
@@ -1,10 +1,11 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net8.0;net462</TargetFrameworks>
     <Nullable>enable</Nullable>
     <IsPackable>false</IsPackable>
     <IsTestProject>true</IsTestProject>
     <ImplicitUsings>disable</ImplicitUsings>
+    <LangVersion>12</LangVersion>
```
stephentoub: Why 12?

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -7,7 +8,6 @@
 using System.Text.Json.Nodes;
 using System.Threading;
 using System.Threading.Tasks;
-using Anthropic;
```
stephentoub, line 10: Did you run the format script? I'd previously had Anthropic at the top and it moved it down.

stephentoub:
Based on https://github.com/anthropics/anthropic-sdk-csharp/issues/27, are tests failing now?

src/Anthropic.Tests/Anthropic.Tests.csproj
rogerbarreto: Given that the latest LTS SDK version this project is being compiled with is 8.0 I would keep it the 12.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-versioning

When this repo shifts to .NET 10 SDK and target, lang should be upgraded to 14.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
rogerbarreto: Yeah, just did, didn't complain about the order.

src/Anthropic.Tests/Anthropic.Tests.csproj
stephentoub: > Given that the latest LTS SDK version this project is being compiled with is 8.0 I would keep it the 12.

The src is building for net9.0, which maps to LangVersion 13.
https://github.com/anthropics/anthropic-sdk-csharp/blob/ff89a7867a520ff8a880d5ae0b65aa19f5127aac/src/Anthropic/Anthropic.csproj#L10

rogerbarreto:
> Based on #27, are tests failing now?

All are.

image

TomerAberbach:
I'm working on support for .NET Standard 2.0 in codegen and am looking at testing (and therefore looking at this PR!)

One thing: I would prefer `./scripts/test` to run _all_ tests, including for .NET Standard 2.0, since that's the expectation for running `./scripts/test`. Can we make it run for both frameworks? ([could also be in parallel](https://stackoverflow.com/a/5553774))

We can still have the CI jobs split up and just use the `dotnet` CLI directly, but it'd be ideal that someone developing locally sees all failing tests when running `./scripts/test`

TomerAberbach:
Oh hmmm, looks like just adding the target frameworks and running `dotnet test`  already tries to test all the frameworks, so do we even need another CI job or any special logic?

TomerAberbach:
You might actually just wanna wait, since my codegen change will be merged soon anyway

TomerAberbach:
I am, however, running your CI cause I'm curious if/how you're getting `net462` to run (since it needs mono, I think?)

EDIT: I see you're running windows, makes sense :)

TomerAberbach:
Okay, we now have test coverage and everything passing as of my latest codegen change. It will be in the release PR soon, so closing this PR

PR #48, opened 2025-12-01
Use JsonSerializer.DeserializeAsync when parsing response stream

stephentoub:
It should use async I/O rather than sync I/O when reading the HTTP response stream.

### DIFF
--- a/src/Anthropic/AnthropicClient.cs
+++ b/src/Anthropic/AnthropicClient.cs
@@ -126,7 +126,7 @@ public class AnthropicClient : IAnthropicClient
                 {
                     throw AnthropicExceptionFactory.CreateApiException(
                         response.Message.StatusCode,
-                        await response.ReadAsString(cancellationToken).ConfigureAwait(false)
+                        await response.ReadAsStringAsync(cancellationToken).ConfigureAwait(false)
                     );
                 }
                 catch (HttpRequestException e)
--- a/src/Anthropic/Core/HttpResponse.cs
+++ b/src/Anthropic/Core/HttpResponse.cs
@@ -14,7 +14,7 @@ public sealed class HttpResponse : IDisposable
 
     public CancellationToken CancellationToken { get; init; } = default;
 
-    public async Task<T> Deserialize<T>(CancellationToken cancellationToken = default)
+    public async Task<T> DeserializeAsync<T>(CancellationToken cancellationToken = default)
     {
         using var cts = CancellationTokenSource.CreateLinkedTokenSource(
             this.CancellationToken,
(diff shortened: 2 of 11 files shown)


TomerAberbach:
Good catch! But I think we should make this change in codegen instead

I also don't think we should change the names of these methods to be suffixed, since they're all async. In my mind it only makes sense to suffix if we'd have both sync and sync variants (and btw, if we did that in this PR, then it'd result in compilation errors the next time we codegen a new service method, and it'd still use the original names)

stephentoub:
> But I think we should make this change in codegen instead

Ok. Feel free to close this of course.

> it only makes sense to suffix if we'd have both sync and sync variants

That's not actually the guidance for the .NET ecosystem. In general, developers expect that {Value}Task-returning methods are suffixed with "Async". Analyzers like https://microsoft.github.io/vs-threading/analyzers/VSTHRD200.html and https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1046/ exist to help enforce that. If you look at the core .NET libraries, APIs that only expose async methods use Async suffixes, e.g. `ChannelReader` (https://learn.microsoft.com/en-us/dotnet/api/system.threading.channels.channelreader-1) has "ReadAsync", "ReadAllAsync", and "WaitToReadAsync", not "Read", "ReadAll", and "WaitToRead". HttpClient has "GetAsync" but no "Get". Etc.

TomerAberbach:
> Ok. Feel free to close this of course.

👍 

> That's not actually the guidance for the .NET ecosystem. In general, developers expect that {Value}Task-returning methods are suffixed with "Async". Analyzers like [microsoft.github.io/vs-threading/analyzers/VSTHRD200.html](https://microsoft.github.io/vs-threading/analyzers/VSTHRD200.html) and [josefpihrt.github.io/docs/roslynator/analyzers/RCS1046](https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1046/) exist to help enforce that. If you look at the core .NET libraries, APIs that only expose async methods use Async suffixes, e.g. `ChannelReader` ([learn.microsoft.com/en-us/dotnet/api/system.threading.channels.channelreader-1](https://learn.microsoft.com/en-us/dotnet/api/system.threading.channels.channelreader-1)) has "ReadAsync", "ReadAllAsync", and "WaitToReadAsync", not "Read", "ReadAll", and "WaitToRead". HttpClient has "GetAsync" but no "Get". Etc.

I hear you, and I think the naming scheme of the core .NET libraries makes sense because they have many methods and it's convenient to immediately know whether a method returns an awaitable thing, but I think it would be verbose and redundant for an SDK, where essentially everything is async.

This is a good datapoint for us to consider this change in the future though. Out of curiosity, would you have expected just these methods you renamed to be `Async`-suffixed or would you have also expected all the service class methods to be `Async`-suffixed?

stephentoub:
> Out of curiosity, would you have expected just these methods you renamed to be Async-suffixed or would you have also expected all the service class methods to be Async-suffixed?

All :-) or more specifically anything that returns a Task or ValueTask.

TomerAberbach:
Alright, well, I'll take your opinion as another data point :)

I just merged a codegen change for using `DeserializeAsync` so that should appear in the release PR soon.

PR #50, opened 2025-12-01
Upgrade to .NET 10

stephentoub:
- Use .NET 10 SDK
- Use C# 14
- Add net10.0 asset to nuget packages
- Change samples to use .NET 10

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -9,8 +9,8 @@
         // For the Prism mock server.
         "ghcr.io/devcontainers/features/node:1": {},
         "ghcr.io/devcontainers/features/dotnet:2": {
-            "version": "8.0",
-            "additionalVersions": "9.0"
+            "version": "10.0",
+            "additionalVersions": "9.0, 8.0"
         }
     }
 }
--- a/global.json
+++ b/global.json
@@ -1,0 +1,6 @@
+{
+  "sdk": {
+    "version": "10.0.100",
+    "rollForward": "minor"
+  }
+}
--- a/src/Anthropic.Foundry/Anthropic.Foundry.csproj
+++ b/src/Anthropic.Foundry/Anthropic.Foundry.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net9.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net10.0;net9.0;net8.0;netstandard2.0</TargetFrameworks>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
     <RootNamespace>Anthropic.Foundry</RootNamespace>
@@ -13,7 +13,6 @@
     <PackageReference Include="Azure.Identity" Version="1.17.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.6.3" />
   </ItemGroup>
   <PropertyGroup>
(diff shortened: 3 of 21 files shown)


src/Anthropic.Foundry/Anthropic.Foundry.csproj
```diff
@@ -13,7 +13,6 @@
     <PackageReference Include="Azure.Identity" Version="1.17.0" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
```
TomerAberbach, line 17: Why remove this?

global.json
TomerAberbach, line 1: What is this for?

src/Anthropic.Foundry/Anthropic.Foundry.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net9.0;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net10.0;net9.0;net8.0;netstandard2.0</TargetFrameworks>
```
TomerAberbach, line 3: Do we really need to compile 4 different versions? What's the thinking around which versions we should keep over time?

src/Anthropic.Tests/Anthropic.Tests.csproj
```diff
@@ -1,12 +1,12 @@
...
     <IsTestProject>true</IsTestProject>
     <IsPackable>false</IsPackable>
 
-    <!-- Run tests on net8.0, which will use the SDK's net8.0 DLL, and on net462, which will use the
+    <!-- Run tests on net10.0, which will use the SDK's net10.0 DLL, and on net462, which will use the
     SDK's netstandard2.0 DLL. -->
-    <TargetFrameworks>net8.0;net462</TargetFrameworks>
-    <LangVersion>12</LangVersion>
+    <TargetFrameworks>net10.0;net462</TargetFrameworks>
+    <LangVersion>latest</LangVersion>
```
TomerAberbach, line 9: Why the change to latest?

rogerbarreto (approved):

.devcontainer/devcontainer.json
```diff
@@ -9,7 +9,7 @@
         // For the Prism mock server.
         "ghcr.io/devcontainers/features/node:1": {},
         "ghcr.io/devcontainers/features/dotnet:2": {
-            "version": "8.0",
+            "version": "10.0",
             "additionalVersions": "9.0"
```
TomerAberbach: If we keep all three versions in compilation, then this should have all those

src/Anthropic.Foundry/Anthropic.Foundry.csproj
```diff
@@ -27,14 +26,14 @@
     <PackageIcon>logo.png</PackageIcon>
     <RepositoryUrl>https://github.com/anthropics/anthropic-sdk-csharp</RepositoryUrl>
     <RepositoryType>git</RepositoryType>
-    <LangVersion>Latest</LangVersion>
+    <LangVersion>latest</LangVersion>
```
TomerAberbach, line 29: Assuming we don't use `latest` in the other files, can we switch this to be the same? I think we missed it after changing from `latest` in codegen (since it's custom code)

src/Anthropic.Foundry/Anthropic.Foundry.csproj
stephentoub: It's not necessary. Any reliance on it in this assembly is inherited from the dependency on Anthropic.

(Separately, I think the use of it is questionable. It looks like it's being used for FrozenDictionary, in support of freezing collections to make them read-only, but FrozenDictionary adds additional overheads, as it's intended to be used in scenarios where it'll be created once and then used many, many times, which is a different use case. I think the use in this library would be better off just using ReadOnlyDictionary or the like.)

global.json
stephentoub: It specifies what version of the .NET SDK should be used. Specifying it makes the experience more consistent, as anyone coming to the project will be guaranteed to be using an SDK in the specified range (in this case, a .NET 10 SDK). When .NET 11 previews come out, for example, the .NET 11 SDK will have different defaults, like LangVersion=15, so settings like "latest" for LangVersion will end up being inconsistent across folks working in the project without such a global.json.

I didn't do so in this PR, but it also enables consolidating version details. Right now, there are multiple occurrences of .NET version numbers in the various .github yml files. Those can instead be changed to say "just use whatever is in the global.json", so that the next time we want to make a change, we only need to change the global.json and not all of those ymls.

src/Anthropic.Foundry/Anthropic.Foundry.csproj
stephentoub: There are at most three versions of .NET core in support at any point in time. The typical answer is to have an asset for each supported version, plus netstandard2.0 (and possibly some .NET Framework specific version). When net11.0 comes out, net8.0 and net9.0 will be out of support, so this would change in Nov 2026 to just being net11.0;net10.0;netstandard2.0.

src/Anthropic.Tests/Anthropic.Tests.csproj
stephentoub: C# 12 is a downgrade from net10.0's default, which is C# 14. If we want to be specific, we could say "14" instead of "latest". The pro to "latest" is that it'll stay up-to-date with the SDK being used. The con is it can change based on what SDK is being used, which is an example of why having a global.json is valuable.

src/Anthropic.Foundry/Anthropic.Foundry.csproj
stephentoub: Sorry, I didn't follow the comment... what do you want to change this to be?

stephentoub:
Seems like this isn't desired right now and it's hard to keep up-to-date, so I'll close it.

PR #55, opened 2025-12-02
Update Anthropic Foundry Identity Token Credential

rogerbarreto:
# Problem Statement

### Problem Statement

The current implementation accepts only a static `AccessToken`, which is suitable for scenarios like an `ApiKey` that remains valid indefinitely during the client's lifetime. However, this approach fails for other `AccessToken` types that expire and require periodic renewal.

To address this, the change delegates token management to the specialized `TokenCredential` implementation.


Anthropic.sln
```diff
@@ -11,6 +11,14 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "src", "src", "{827E0CD3-B72
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Anthropic.Foundry", "src\Anthropic.Foundry\Anthropic.Foundry.csproj", "{DD0E539D-6D5F-45EB-A807-01BE0A443604}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{93E58BA5-CEFE-447E-AC0C-F2C5BC4C411D}"
```
rogerbarreto, line 14: FYI, this adds some items from the root folder to be visible in the solution explorer.

src/Anthropic.Foundry/AnthropicFoundryIdentityTokenCredentials.cs
```diff
@@ -7,32 +8,57 @@ namespace Anthropic.Foundry;
...
     {
-        if (apiKey.Equals(default) || string.IsNullOrWhiteSpace(apiKey.Token))
+        if (apiKey.Equals(default) || string.IsNullOrWhiteSpace(apiKey.Token) || apiKey.ExpiresOn <= DateTimeOffset.Now)
         {
             throw new ArgumentNullException(
                 nameof(apiKey),
-                "Invalid/Empty api key struct is not valid"
+                "Invalid/Empty or already expired api key struct is not valid"
             );
         }
```
sd-st: Since this SDK is still in beta I think we should probably just remove it instead of marking it deprecated. I'm not too concerned about it being a breaking change

src/Anthropic.Foundry/AnthropicFoundryIdentityTokenCredentials.cs
```diff
@@ -7,32 +8,57 @@ namespace Anthropic.Foundry;
...
-    public string ResourceName { get; }
-
     public void Apply(HttpRequestMessage requestMessage)
     {
-        //TODO implement token refresh
-        requestMessage.Headers.Authorization =
-            new System.Net.Http.Headers.AuthenticationHeaderValue("bearer", _apiKey.Token);
+        lock (this._tokenLock)
+        {
+            if (this._cachedAccessToken is null || this._cachedAccessToken.Value.ExpiresOn <= DateTimeOffset.Now)
```
sd-st: Rather than just checking the date ourselves, we probably want to refresh the token if we get an expire error from the server as well. Otherwise there's an opportunity for a race where the token expires after we check but before we make the request (or the computer's clock isn't synced, etc)

src/Anthropic/AnthropicClient.cs
```diff
@@ -291,11 +291,9 @@ static bool ShouldRetry(HttpResponse response)
             // Retry on lock timeouts
             409
             or
-#if !NETSTANDARD2_0_OR_GREATER
```
sd-st, line 410: Good catch, I guess this was from a bad merge. I checked and we're not code-genning this

.editorconfig
```diff
@@ -23,6 +23,7 @@ insert_final_newline = true
 
 dotnet_diagnostic.CA1510.severity = none  # Use ArgumentNullException throw helper
 dotnet_diagnostic.IDE0060.severity = none # Caused by resource with no methods and no subresources
+dotnet_diagnostic.IDE0009.severity = none # Caused by resource with no methods and no subresources
```
sd-st: Just curious, what is this from? Generally I would prefer we fix rather than ignore but I'm open to hearing your thoughts. We'll want to change the comment if we keep it though

Anthropic.sln
sd-st: I must admit I am not a Visual Studio enjoyer - the last version I used regularly was VS 2010. Do the files just not show up if they're not in here? I'm thinking I will probably backport this to codegen, good catch

sd-st (changes requested):
Thanks for your PR! Just had a few minor comments

src/Anthropic.Foundry/AnthropicFoundryIdentityTokenCredentials.cs
rogerbarreto: Normally this will happen from the resilience retrial itself, so basically if the token is expired, the server will return 403 or 401, then a new request will be made, and this new request will call the Token, which in this second trip should give the refreshed token.

> or the computer's clock isn't synced, etc)

That's a good point, if we can't thrust the ExpiresOn, we might just cache the Token from a specific minimum default amount of time (to prevent round-trips)

.editorconfig
rogerbarreto: I rather fix as well but my IDE has auto formatting and some default severity checks for solutions.

When no rule is defined my IDE assumes is an issue and start formatting things, like `this` prefix into instance files which is the `severity = none` update about.

I updated the description which was wrong, should be:

`Add 'this' or 'Me' qualification`

src/Anthropic.Foundry/AnthropicFoundryIdentityTokenCredentials.cs
rogerbarreto: Updated the caching logic to have a fixed cache timeout, defaulting to 5 min.

.editorconfig
sd-st: Does this come from your code or from existing code? If the former we should probably just add the qualifiers, and if the latter I will see if we want to backport this to codegen

sd-st:
Would you be able to retarget this to the `next` branch?

rogerbarreto:
> Would you be able to retarget this to the next branch?

Sure, done!

.editorconfig
rogerbarreto: @sd-st This comes from my IDE formatting environment, which are good practices nonetheless, but I understand shouldn't be imposable over repositories that have no defined rules.

Anthropic.sln
rogerbarreto: > I must admit I am not a Visual Studio enjoyer - the last version I used regularly was VS 2010. Do the files just not show up if they're not in here? I'm thinking I will probably backport this to codegen, good catch

Yeah, I don't like this very much either, surprisingly VS still don't show already known files automatically into solution (all folders in Solution Explorer are virtual structures), and we need to force it to see things in this view that in theory it should already know.

src/Anthropic.Foundry/AnthropicFoundryIdentityTokenCredentials.cs
sd-st: > Normally this will happen from the resilience retrial itself, so basically if the token is expired, the server will return 403 or 401, then a new request will be made, and this new request will call the Token, which in this second trip should give the refreshed token.

Hmm I don't think I fully follow? It seems like even if the request fails with a 401/403 there's no mechanism which tells the `AnthropicFoundryIdentityTokenCredentials` that it needs to flush its cache and get a new token. Otherwise if the cached token is expired I think we will just keep using it, no?

rogerbarreto: Removing the Caching and any other complexity from this implementation. 

Deferring all responsibility to the `TokenCredential` implementation as this is by design.

Using `DefaultAzureCredential` has already all the refresh / caching behavior by default, clear stated in the documentation.

> [!NOTE]
> Token lifetime and refreshing is handled automatically
> See: https://github.com/Azure/azure-sdk-for-net/blob/main/sdk/identity/Azure.Identity/src/Credentials/DefaultAzureCredential.cs#L121-L126

sd-st:
@rogerbarreto There's a lint failure :disappointed:

rogerbarreto:
Something is off.

This is what I get from my current branch, up-to-date, lint and format don't require any change.

image

rogerbarreto:
@sd-st, merged `next` into the PR to see if a new pipeline run can work.

sd-st:
Hey @rogerbarreto do you mind rebasing this? Looks like something got messed up and the diff is huge now.

rogerbarreto:
@sd-st, Done (from what I could remember), I also noticed with the latest `next` branch, a bunch of Unit Tests are failing... unrelated to this PR. 

> [!NOTE]
> The process of the automation / codegen you are using in the repo is also messing up with all the PR's and possibly touching/altering / force-pushing into PRs git history is the reason. 
> Ideally for a healthy contribution environment this should be avoided.

sd-st:
Looks like there's a lint failure from the formatting

sd-st (approved):
Thank you!

JPVenson:
this breaks the build as the method is now still declared async but has no await

Issue #45, opened 2025-11-26
Missing `code_execution-2025-08-25` beta header prevents use of non-legacy Code Execution tool


PederHP:
## Title
**Missing `code_execution-2025-08-25` beta header prevents use of new Code Execution tool**

### Problem
The `AnthropicBeta` enum currently only includes the legacy `code_execution-2025-05-22` beta header value. The newer `code_execution-2025-08-25` header is missing, which prevents users from enabling the updated Code Execution server-side tool.

### Impact
- **Microsoft.Extensions.AI users**: The `HostedCodeInterpreterTool` in the MEAI extensions maps to the new code execution tool, but there's no way to pass the required beta header through the SDK.
- **Direct SDK users**: Anyone wanting to use the latest code execution tool capabilities cannot do so, as the API rejects requests without the correct beta header.

Attempting to use the new code execution tool results in an API error since the legacy header only enables the legacy tool variant.

### Expected Behavior
The SDK should expose `CodeExecution2025_08_25` (or equivalent) in the `AnthropicBeta` enum so users can opt into the updated code execution functionality.

### Reference
- [[Code Execution Tool Documentation](https://platform.claude.com/docs/en/agents-and-tools/tool-use/code-execution-tool)](https://platform.claude.com/docs/en/agents-and-tools/tool-use/code-execution-tool)

PR #58, opened 2025-12-03
Idiomatic Skills support for MEAI via AsAITool() extension

PederHP:
Added an `AsAITool` extension to `BetaSkillParam` to avoid having to use `RawRepresentationFactory` when using skills with the Microsoft.Extensions.AI extensions. Had to use a string for one beta header due to #45.

Example usage:
```csharp
var options = new ChatOptions
{
    ModelId = "claude-sonnet-4-5-20250929",
    MaxOutputTokens = 4096,
    Tools =
    [
        new BetaSkillParams { Type = BetaSkillParamsType.Anthropic, SkillID = "pptx", Version = "latest" }.AsAITool(),
        new BetaSkillParams { Type = BetaSkillParamsType.Anthropic, SkillID = "xlsx", Version = "latest" }.AsAITool(),
    ]
};

```


TomerAberbach:
Thanks :)

@stephentoub mind taking a look and providing your thoughts?

PederHP:
@stephentoub Would it make sense to create a `HostedSkillsTool` to make this even more convenient? I started making a PR for `Microsoft.Extensions.AI` for this, but it seemed to be a very Claude-specific abstraction, so I ended up not submitting it. But maybe it would make sense to have in the SDK? I don't know if the maintainers would want to add MEAI specific types to the SDK. Another option could be to extend `HostedCodeInterpreterTool` in MEAI with something that can map to skills, but that feels like polluting the abstraction.

stephentoub:
How about an `AsAITool` extension method for `BetaSkill` (or `BetaSkillParam`... I'm not sure what the difference is), ala https://github.com/anthropics/anthropic-sdk-csharp/blob/957b674e9bc2db9e9526a402bb2338778788413e/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs#L76 ?

PederHP:
> How about an `AsAITool` extension method for `BetaSkillParam`, ala
> 
> https://github.com/anthropics/anthropic-sdk-csharp/blob/957b674e9bc2db9e9526a402bb2338778788413e/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs#L76
> 
> ?

That's a pretty great idea! We'd still need a specialized AITool type to map to the non-tool Container property and the Skills in it, but then it can be private.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -19,6 +20,85 @@ namespace Microsoft.Extensions.AI;
...
+                    ?? throw new InvalidOperationException(
+                        "MaxOutputTokens must be specified in ChatOptions when using WithSkills."
+                    ),
+                Messages = [],
+                Container = new BetaContainerParams() { Skills = skills },
+                // Required beta headers for skills
+                Betas =
+                [
+                    "code-execution-2025-08-25", // Enables code execution (required for Skills)
+                    AnthropicBeta.Skills2025_10_02, // Enables Skills API
```
stephentoub: Separate from this PR, currently the IChatClient implementation doesn't automatically opt-into any of these... should it? i.e. should it try to detect which features are being used and ensure the appropriate feature flags are included? Should it always include all of them?

PederHP: I think at least some of them should be auto-include. Currently, I need to have code like this on the application side:

```csharp
var betas = new List<ApiEnum<string, AnthropicBeta>>();
if (hasExtendedThinking)
{
    betas.Add(AnthropicBeta.InterleavedThinking2025_05_14);
}
if (hasCodeExecution)
{
    betas = [.. betas, "code-execution-2025-08-25"];
}
```` 

I think the CodeExecution beta headers should be auto-include for the related tools. Interleaved thinking I'm not sure about. I can't see why someone would not want it (if they're using thinking), but someone out there might have a reason not to. I had honestly expected that one to be out of beta by now.

File: src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs [hunk 2]
```diff
@@ -19,6 +20,72 @@ namespace Microsoft.Extensions.AI;
...
+    /// <summary>
+    /// An <see cref="AITool"/> that wraps a <see cref="BetaSkillParams"/> for use with the Anthropic Skills API.
+    /// </summary>
+    /// <remarks>
+    /// This tool is used internally to carry skill configuration through the MEAI tools pipeline.
+    /// When detected by the <see cref="IChatClient"/> returned by <see cref="AsIChatClient"/>,
+    /// it will automatically configure the container with skills, add required beta headers,
+    /// and ensure the code execution tool is present.
+    /// </remarks>
+    internal sealed class HostedSkillTool(BetaSkillParams skillParams) : AITool
```
stephentoub: I suggest:
- Making this private rather than internal
- Move this down next to ToolUnionAITool at the bottom of the file
- Naming this BetaSkillParamsAITool

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -19,6 +20,72 @@ namespace Microsoft.Extensions.AI;
...
+    {
+        /// <summary>
+        /// The parameters defining the skill.
+        /// </summary>
+        public BetaSkillParams SkillParams { get; } = skillParams;
+
+        /// <inheritdoc />
+        public override string Name => SkillParams.SkillID;
```
stephentoub: ToolUnionAITool also overrides GetService to support returning the underlying BetaSkillsParams. I suggest this does the same thing.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1000,6 +1072,55 @@ is true
...
+                        // Ensure code execution tool is present
+                        bool hasCodeExecutionTool =
+                            createdTools?.Any(t => t.Value is BetaCodeExecutionTool20250825)
+                            == true;
+                        if (!hasCodeExecutionTool)
+                        {
+                            (createdTools ??= []).Add(new BetaCodeExecutionTool20250825());
+                            createParams = createParams with { Tools = createdTools };
+                        }
+                    }
```
stephentoub: We'll want to unify this with what's being done in https://github.com/anthropics/anthropic-sdk-csharp/pull/63/files#diff-3d7ece0795ffe039e594b56c7c3734c21036cde6a45e168a9a20157e0a448239R977. This whole section should consolidate down to just a couple of lines.

PederHP: Ah yes, good point. That does consolidate it a lot. The merging with existing skills in the container takes up a fair bit of the consolidated code. We can't really avoid it, even if it is an edge case (mixing RawRepresentationFactory and AsAITool), but maybe there is a more elegant/concise way to merge them.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -886,10 +938,17 @@ createParams.OutputFormat is null
...
                     List<BetaRequestMCPServerURLDefinition>? mcpServers =
                         createParams.MCPServers?.ToList();
+                    List<BetaSkillParams>? skills = null;
                     foreach (var tool in tools)
                     {
                         switch (tool)
                         {
+                            case BetaSkillsParamsAITool skillTool:
+                                (betaHeaders ??= []).Add("skills-2025-10-02");
+                                (betaHeaders ??= []).Add("code-execution-2025-08-25");
```
stephentoub:
```suggestion
                                betaHeaders.Add("code-execution-2025-08-25");

```

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -991,6 +1052,36 @@ is true
...
+
+                        createParams = createParams with
+                        {
+                            Container = new BetaContainerParams() { Skills = skills },
+                        };
+
+                        // Ensure code execution tool is present
+                        bool hasCodeExecutionTool =
+                            createdTools?.Any(t => t.Value is BetaCodeExecutionTool20250825)
+                            == true;
```
stephentoub, line 1080: What happens if the same tool is added multiple times? I'm wondering if this is actually necessary.

stephentoub (approved):

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: Some APIs give an error. I think the Anthropic one might be one of them. Tokens are also wasted, if it is allowed.

sd-st (approved):
Thanks!

sd-st:
Could you target against `next` instead of `main`? We generally only merge into `main` for a release

stephentoub (approved):

sd-st:
Thanks again!

PR #61, opened 2025-12-03
Add Bedrock as provider for SDK

JPVenson:
Adds Bedrock implementation for usage with api.

Notes:
 Only the Model invoke and streaming endpoint is currently supported

### DIFF
--- a/src/Anthropic.Tests/Models/Beta/Messages/Batches/BatchCreateParamsTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/Batches/BatchCreateParamsTest.cs
@@ -966,6 +966,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedModel, model.Model);
         Assert.Equal(expectedContainer, model.Container);
         Assert.Equal(expectedContextManagement, model.ContextManagement);
+        Assert.NotNull(model.MCPServers);
         Assert.Equal(expectedMCPServers.Count, model.MCPServers.Count);
         for (int i = 0; i < expectedMCPServers.Count; i++)
         {
@@ -975,6 +976,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedOutputConfig, model.OutputConfig);
         Assert.Equal(expectedOutputFormat, model.OutputFormat);
         Assert.Equal(expectedServiceTier, model.ServiceTier);
+        Assert.NotNull(model.StopSequences);
         Assert.Equal(expectedStopSequences.Count, model.StopSequences.Count);
         for (int i = 0; i < expectedStopSequences.Count; i++)
         {
@@ -985,6 +987,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedTemperature, model.Temperature);
         Assert.Equal(expectedThinking, model.Thinking);
         Assert.Equal(expectedToolChoice, model.ToolChoice);
+        Assert.NotNull(model.Tools);
         Assert.Equal(expectedTools.Count, model.Tools.Count);
         for (int i = 0; i < expectedTools.Count; i++)
         {
@@ -1360,6 +1363,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedModel, deserialized.Model);
         Assert.Equal(expectedContainer, deserialized.Container);
         Assert.Equal(expectedContextManagement, deserialized.ContextManagement);
+        Assert.NotNull(deserialized.MCPServers);
         Assert.Equal(expectedMCPServers.Count, deserialized.MCPServers.Count);
         for (int i = 0; i < expectedMCPServers.Count; i++)
         {
@@ -1369,6 +1373,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedOutputConfig, deserialized.OutputConfig);
         Assert.Equal(expectedOutputFormat, deserialized.OutputFormat);
         Assert.Equal(expectedServiceTier, deserialized.ServiceTier);
+        Assert.NotNull(deserialized.StopSequences);
         Assert.Equal(expectedStopSequences.Count, deserialized.StopSequences.Count);
         for (int i = 0; i < expectedStopSequences.Count; i++)
         {
@@ -1379,6 +1384,7 @@ public class ParamsTest : TestBase
         Assert.Equal(expectedTemperature, deserialized.Temperature);
         Assert.Equal(expectedThinking, deserialized.Thinking);
         Assert.Equal(expectedToolChoice, deserialized.ToolChoice);
+        Assert.NotNull(deserialized.Tools);
         Assert.Equal(expectedTools.Count, deserialized.Tools.Count);
         for (int i = 0; i < expectedTools.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaClearToolUses20250919EditTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaClearToolUses20250919EditTest.cs
@@ -30,6 +30,7 @@ public class BetaClearToolUses20250919EditTest : TestBase
         Assert.True(JsonElement.DeepEquals(expectedType, model.Type));
         Assert.Equal(expectedClearAtLeast, model.ClearAtLeast);
         Assert.Equal(expectedClearToolInputs, model.ClearToolInputs);
+        Assert.NotNull(model.ExcludeTools);
         Assert.Equal(expectedExcludeTools.Count, model.ExcludeTools.Count);
         for (int i = 0; i < expectedExcludeTools.Count; i++)
         {
@@ -85,6 +86,7 @@ public class BetaClearToolUses20250919EditTest : TestBase
         Assert.True(JsonElement.DeepEquals(expectedType, deserialized.Type));
         Assert.Equal(expectedClearAtLeast, deserialized.ClearAtLeast);
         Assert.Equal(expectedClearToolInputs, deserialized.ClearToolInputs);
+        Assert.NotNull(deserialized.ExcludeTools);
         Assert.Equal(expectedExcludeTools.Count, deserialized.ExcludeTools.Count);
         for (int i = 0; i < expectedExcludeTools.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaCodeExecutionTool20250522Test.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaCodeExecutionTool20250522Test.cs
@@ -30,6 +30,7 @@ public class BetaCodeExecutionTool20250522Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, model.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, model.Type));
+        Assert.NotNull(model.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, model.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
@@ -83,6 +84,7 @@ public class BetaCodeExecutionTool20250522Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, deserialized.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, deserialized.Type));
+        Assert.NotNull(deserialized.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, deserialized.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaCodeExecutionTool20250825Test.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaCodeExecutionTool20250825Test.cs
@@ -33,6 +33,7 @@ public class BetaCodeExecutionTool20250825Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, model.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, model.Type));
+        Assert.NotNull(model.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, model.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
@@ -89,6 +90,7 @@ public class BetaCodeExecutionTool20250825Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, deserialized.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, deserialized.Type));
+        Assert.NotNull(deserialized.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, deserialized.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaContainerParamsTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaContainerParamsTest.cs
@@ -35,6 +35,7 @@ public class BetaContainerParamsTest : TestBase
         ];
 
         Assert.Equal(expectedID, model.ID);
+        Assert.NotNull(model.Skills);
         Assert.Equal(expectedSkills.Count, model.Skills.Count);
         for (int i = 0; i < expectedSkills.Count; i++)
         {
@@ -98,6 +99,7 @@ public class BetaContainerParamsTest : TestBase
         ];
 
         Assert.Equal(expectedID, deserialized.ID);
+        Assert.NotNull(deserialized.Skills);
         Assert.Equal(expectedSkills.Count, deserialized.Skills.Count);
         for (int i = 0; i < expectedSkills.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaContainerTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaContainerTest.cs
@@ -39,6 +39,7 @@ public class BetaContainerTest : TestBase
 
         Assert.Equal(expectedID, model.ID);
         Assert.Equal(expectedExpiresAt, model.ExpiresAt);
+        Assert.NotNull(model.Skills);
         Assert.Equal(expectedSkills.Count, model.Skills.Count);
         for (int i = 0; i < expectedSkills.Count; i++)
         {
@@ -106,6 +107,7 @@ public class BetaContainerTest : TestBase
 
         Assert.Equal(expectedID, deserialized.ID);
         Assert.Equal(expectedExpiresAt, deserialized.ExpiresAt);
+        Assert.NotNull(deserialized.Skills);
         Assert.Equal(expectedSkills.Count, deserialized.Skills.Count);
         for (int i = 0; i < expectedSkills.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaContextManagementConfigTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaContextManagementConfigTest.cs
@@ -36,6 +36,7 @@ public class BetaContextManagementConfigTest : TestBase
             },
         ];
 
+        Assert.NotNull(model.Edits);
         Assert.Equal(expectedEdits.Count, model.Edits.Count);
         for (int i = 0; i < expectedEdits.Count; i++)
         {
@@ -101,6 +102,7 @@ public class BetaContextManagementConfigTest : TestBase
             },
         ];
 
+        Assert.NotNull(deserialized.Edits);
         Assert.Equal(expectedEdits.Count, deserialized.Edits.Count);
         for (int i = 0; i < expectedEdits.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaMemoryTool20250818Test.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaMemoryTool20250818Test.cs
@@ -45,6 +45,7 @@ public class BetaMemoryTool20250818Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, model.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, model.Type));
+        Assert.NotNull(model.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, model.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
@@ -52,6 +53,7 @@ public class BetaMemoryTool20250818Test : TestBase
         }
         Assert.Equal(expectedCacheControl, model.CacheControl);
         Assert.Equal(expectedDeferLoading, model.DeferLoading);
+        Assert.NotNull(model.InputExamples);
         Assert.Equal(expectedInputExamples.Count, model.InputExamples.Count);
         for (int i = 0; i < expectedInputExamples.Count; i++)
         {
@@ -131,6 +133,7 @@ public class BetaMemoryTool20250818Test : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedName, deserialized.Name));
         Assert.True(JsonElement.DeepEquals(expectedType, deserialized.Type));
+        Assert.NotNull(deserialized.AllowedCallers);
         Assert.Equal(expectedAllowedCallers.Count, deserialized.AllowedCallers.Count);
         for (int i = 0; i < expectedAllowedCallers.Count; i++)
         {
@@ -138,6 +141,7 @@ public class BetaMemoryTool20250818Test : TestBase
         }
         Assert.Equal(expectedCacheControl, deserialized.CacheControl);
         Assert.Equal(expectedDeferLoading, deserialized.DeferLoading);
+        Assert.NotNull(deserialized.InputExamples);
         Assert.Equal(expectedInputExamples.Count, deserialized.InputExamples.Count);
         for (int i = 0; i < expectedInputExamples.Count; i++)
         {
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaMemoryTool20250818ViewCommandTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaMemoryTool20250818ViewCommandTest.cs
@@ -21,6 +21,7 @@ public class BetaMemoryTool20250818ViewCommandTest : TestBase
 
         Assert.True(JsonElement.DeepEquals(expectedCommand, model.Command));
         Assert.Equal(expectedPath, model.Path);
+        Assert.NotNull(model.ViewRange);
         Assert.Equal(expectedViewRange.Count, model.ViewRange.Count);
         for (int i = 0; i < expectedViewRange.Count; i++)
         {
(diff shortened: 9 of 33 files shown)


src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -0,0 +1,22 @@
...
+  <PropertyGroup>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <LangVersion>13.0</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <PackageId>Anthropic.Bedrock</PackageId>
+    <VersionPrefix>0.1.0</VersionPrefix>
```
sd-st: I believe we need this version to be the same as the other package versions in this repo as a side effect of our release flow. @TomerAberbach can you confirm?

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>net9.0;net8.0</TargetFrameworks>
```
sd-st: I think we also want .net standard 2.0

src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// </summary>
+    /// <value>
+    /// A string representing the bearer token. This value is set privately and can only be modified within the class.
+    /// </value>
+    public string BearerToken { get; private set; }
+
+    /// <summary>
+    /// Gets the AWS region.
+    /// </summary>
+    public string Region { get; private set; }
```
sd-st: NIT: can we make these `private init` if we never modify them?

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -0,0 +1,22 @@
...
+    <PackageId>Anthropic.Bedrock</PackageId>
+    <VersionPrefix>0.1.0</VersionPrefix>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Anthropic\Anthropic.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.Core" Version="4.0.3.3" />
```
sd-st: Just wondering, do we need this? Seems like we're re-implementing everything from scratch; I don't see this used anywhere

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,257 @@
...
+    {
+        _bedrockCredentials =
+            bedrockCredentials ?? throw new ArgumentNullException(nameof(bedrockCredentials));
+        BaseUrl = new Uri($"https://{ServiceName}.{_bedrockCredentials.Region}.amazonaws.com");
+    }
+
+    /// <summary>
+    /// When set to true Streaming endpoints are called async to the caller and are buffered in memory until consumed, when set to false data is requested from the server on demand.
+    /// </summary>
+    public bool AsyncStreaming { get; set; }
```
sd-st: I am on the fence about this; do we need this level of configurability? Seems like something the streaming aggregators could handle if I understand this correctly. What do the other SDKs do?

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,257 @@
...
+            var bodyContent = JsonNode.Parse(
+                await requestMessage.Content!.ReadAsStringAsync().ConfigureAwait(false)
+            );
+
+            if (bodyContent?["model"] == null)
+            {
+                throw new AnthropicInvalidDataException(
+                    "Expected to find property model in request json but found none."
+                );
+            }
```
sd-st: Given that this should already be set from serializing the params, I think we can probably remove this check. If a user manually constructs this without a model we can let it blow up on the server side

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,257 @@
...
+    // <inheritdoc />
+    public override IAnthropicClient WithOptions(Func<ClientOptions, ClientOptions> modifier)
+    {
+        return new AnthropicBedrockClient(_bedrockCredentials, modifier(this._options));
+    }
+
+    // <inheritdoc />
+    protected override async ValueTask BeforeSend<T>(
+        HttpRequest<T> request,
+        HttpRequestMessage requestMessage,
```
sd-st: Hm I am not the biggest fan of using `HttpRequestMessage` here given that we immediately deserialize the JSON (whereas, if we had the `HttpRequest<T>` we'd have access to all of that stuff directly)

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,257 @@
...
+            if (parsedStreamValue)
+            {
+                requestMessage.Headers.TryAddWithoutValidation("Accept-Encoding", "gzip");
+            }
+        }
+
+        await _bedrockCredentials.Apply(requestMessage).ConfigureAwait(false);
+    }
+
+    private static void ValidateRequest(HttpRequestMessage requestMessage)
```
sd-st: I don't think we need this; we can just make the request and let the server do the validation

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,257 @@
...
+
+        // A decoded AWS EventStream message's payload is JSON. It might look like this (abridged):
+        //
+        //   {"bytes":"eyJ0eXBlIjoi...ZXJlIn19","p":"abcdefghijkl"}
+        //
+        // The value of the "bytes" field is a base-64 encoded JSON string (UTF-8). When decoded, it
+        // might look like this:
+        //
+        //   {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"}}
+        //
```
sd-st: dang what are they doing at AWS lol

src/Anthropic.Bedrock/AnthropicBedrockPrivateKeyCredentials.cs
```diff
@@ -0,0 +1,81 @@
...
+            requestMessage.Headers.ToDictionary(
+                e => e.Key,
+                e => string.Join(" ", e.Value),
+                StringComparer.InvariantCultureIgnoreCase
+            ),
+            payloadHash,
+            ApiAccessKey,
+            ApiSecret
+        );
+        // requestMessage.Headers.TryAddWithoutValidation("Host", requestMessage.RequestUri.Host);
```
sd-st: Probably want to remove this comment

src/Anthropic.Bedrock/AsyncSseEventContentWrapper.cs
```diff
@@ -0,0 +1,170 @@
...
+        CancellationToken cancellationToken
+    )
+    {
+        var stream = new LimitedMemoryStream();
+        Task.Run(
+            async () =>
+            {
+                try
+                {
+                    await SerializeToStreamAsync(stream, null).ConfigureAwait(false);
```
sd-st: Should we pass the cancellation token through to this?

src/Anthropic.Bedrock/AsyncSseEventContentWrapper.cs
```diff
@@ -0,0 +1,170 @@
...
+/// This class converts incoming SSE data from the original stream into a format that can be consumed by HTTP clients,
+/// using a custom memory stream to buffer data as it's loaded from the remote server.
+/// </summary>
+/// <remarks>
+/// This wrapper is designed to handle streaming SSE responses, particularly for AWS Bedrock streaming APIs.
+/// It processes SSE messages asynchronously while providing a synchronous read interface for compatibility
+/// with standard HTTP content readers. The class uses a <see cref="LimitedMemoryStream"/> to buffer
+/// data between the writer (SSE processor) and reader (HTTP client) threads.
+/// </remarks>
+internal class AsyncSseEventContentWrapper : HttpContent
```
sd-st: Are we just doing this so we can reuse some of the client logic? This doesn't feel like a super `HttpContent`-shaped thing and it makes me a little nervous

src/Anthropic.Bedrock/AwsSigner.cs
```diff
@@ -0,0 +1,160 @@
...
+
+/// <summary>
+/// AWS Signing helper
+/// </summary>
+/// <remarks>
+/// Taken from: https://github.com/aws-samples/sigv4-signing-examples/blob/main/no-sdk/dotnet/AWSSigner.cs
+/// </remarks>
+public class AWSSigner
+{
+    public static string GetAuthorizationHeader(
```
sd-st: Do we need 2 constructors here or can we just use one and have default values?

src/Anthropic.Bedrock/IAnthropicBedrockCredentials.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    public string Region { get; }
+    public Task Apply(HttpRequestMessage requestMessage);
+
+    public static async ValueTask<IAnthropicBedrockCredentials?> FromEnv()
+    {
+        const string ENV_API_KEY = "AWS_BEARER_TOKEN_BEDROCK";
+        const string ENV_REGION = "AWS_REGION";
+
+        var region = Environment.GetEnvironmentVariable(ENV_REGION); // this is a bit redundant as FallbackRegionFactory checks for the same value too, but its aligned with the behavior of the java sdk
```
sd-st: Hmm, if the fallback region factory also does this I think we probably don't need it. @TomerAberbach thoughts?

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
sd-st: er wait I have found it, it's just under the `Amazon` namespace

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,165 @@
...
+        ReadOnlySpan<byte> data,
+        ReadOnlySpan<byte> checksum
+    )
+    {
+        var dataChecksum = CRC32.ComputeChecksum(data);
+        var reference = BinaryPrimitives.ReadUInt32BigEndian(checksum);
+        return dataChecksum == reference;
+    }
+
+    public class CRC32
```
sd-st: *casually implements CRC32 checksum*

sd-st (changes requested):

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
JPVenson: Yes i saw that too. Sadly that involes a bit more work too

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: I discussed that with tomer already and that flag plus the AsyncSse streamer will be removed

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: this follows the same pattern as Foundry and the java SDK too. I did not wanted to introduce more breaking changes then absolutly nessesary. But i can rework that if you insist.

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: This follows the same pattern as the java SDK and i think it would be better to have explicit validation here in case you are trying to call any other methds then invoke or stream. Otherwise you would just get a non-speaking 404 respose which i doubt anyone would see as _we_ dont implement those endpoints (if they even exist?)

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: it gets worse when you read the SseStreamHelpers ...

this is the 2nd level of madness

src/Anthropic.Bedrock/AsyncSseEventContentWrapper.cs
JPVenson: this will be removed after consulting with tomer

src/Anthropic.Bedrock/AwsSigner.cs
JPVenson: i cannot follow. Those are just methods and i like to keep the method complexity down.

And this is more or less copied from the AWS example repo so i dont want to touch that more then i need to^^

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: ok fun fact. For _whatever_ reason, dotnet only implements crc32 for ARM platforms and the crc32 helper from the AWS nuget package needs a lot of working around for this particular case ... 


so its actually just easier to do it myself xD

JPVenson: leslie-nielsen-hand-sign-nothing-to-see-here-r90u27eqnsugflzb-1470967591

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
TomerAberbach: Nah, it doesn't have to be, but we should be updating https://github.com/anthropics/anthropic-sdk-csharp/blob/main/.release-please-manifest.json and https://github.com/anthropics/anthropic-sdk-csharp/blob/main/release-please-config.json for this new package

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -0,0 +1,31 @@
...
+  </ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Bcl.Memory" Version="10.0.0" />    
+    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
+    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.6.3" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="..\logo.png" Pack="true" PackagePath="\" />
+    <Compile Include="..\Anthropic\Shims.cs" />
+  </ItemGroup>
```
TomerAberbach: Could merge a bunch of these item groups (the ones that aren't netstandard2.0 specific)

src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
```diff
@@ -0,0 +1,48 @@
...
+
+/// <summary>
+/// Provides bearer token-based authentication credentials for accessing Amazon Bedrock Anthropic API.
+/// </summary>
+/// <remarks>
+/// This class implements the <see cref="IAnthropicBedrockCredentials"/> interface to support
+/// authentication using a bearer token and AWS region for Anthropic API requests through Amazon Bedrock.
+/// The bearer token is applied to the Authorization header of HTTP requests.
+/// </remarks>
+public class AnthropicBedrockApiTokenCredentials : IAnthropicBedrockCredentials
```
TomerAberbach: Should probably be `sealed`

src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
```diff
@@ -0,0 +1,48 @@
...
+    /// </summary>
+    /// <value>
+    /// A string representing the bearer token. This value is set privately and can only be modified within the class.
+    /// </value>
+    public string BearerToken { get; private init; }
+
+    /// <summary>
+    /// Gets the AWS region.
+    /// </summary>
+    public string Region { get; private init; }
```
TomerAberbach: Any reason for this to not just use `public required` props with public `init`?

src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
```diff
@@ -0,0 +1,48 @@
...
+
+    // <inheritdoc />
```
TomerAberbach:
```suggestion
    /// <inheritdoc/>
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+using System.Text.Json.Nodes;
+using Anthropic.Core;
+using Anthropic.Exceptions;
+
+namespace Anthropic.Bedrock;
+
+/// <summary>
+/// Provides an Anthropic client implementation for AWS Bedrock integration.
+/// </summary>
+public class AnthropicBedrockClient : AnthropicClient
```
TomerAberbach: Should probably be `sealed`

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+{
+    private const string ServiceName = "bedrock-runtime";
+    private const string AnthropicVersion = "bedrock-2023-05-31";
+    private const string HeaderAnthropicBeta = "anthropic-beta";
+
+    /// <summary>
+    /// The name of the header that identifies the content type for the "payloads" of AWS
+    /// _EventStream_ messages in streaming responses from Bedrock.
```
TomerAberbach: I'm pretty sure the underscores from the Java documentation was for italics, but I'm pretty sure that doesn't work in C# docs, so I'd remove all these underscores

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+    /// </summary>
+    private const string HeaderPayloadContentType = "x-amzn-bedrock-content-type";
+
+    /// <summary>
+    /// The content type for Bedrock responses containing data in the AWS _EventStream_ format.
+    /// The value of the[HEADER_PAYLOAD_CONTENT_TYPE] header identifies the content type of the
```
TomerAberbach: What's going on with `the[HEADER_PAYLOAD_CONTENT_TYPE]`?

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+
+    /// <summary>
+    /// Creates a new Instance of the <see cref="AnthropicBedrockClient"/>.
+    /// </summary>
+    /// <param name="bedrockCredentials">The credential Provider used to authenticate with the AWS Bedrock service.</param>
+    public AnthropicBedrockClient(IAnthropicBedrockCredentials bedrockCredentials)
+        : base()
+    {
+        _bedrockCredentials =
+            bedrockCredentials ?? throw new ArgumentNullException(nameof(bedrockCredentials));
```
TomerAberbach: Is there a point to these null checks when the parameter is annotated as non-nullable? It seems overkill to me. Like if end-users are gonna ignore nullability annotations, then that's on them

(and we're not doing the same check in the constructor below)

Anthropic.sln
```diff
@@ -57,6 +59,18 @@ Global
 		{DD0E539D-6D5F-45EB-A807-01BE0A443604}.Release|x64.Build.0 = Release|Any CPU
 		{DD0E539D-6D5F-45EB-A807-01BE0A443604}.Release|x86.ActiveCfg = Release|Any CPU
 		{DD0E539D-6D5F-45EB-A807-01BE0A443604}.Release|x86.Build.0 = Release|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|x64.ActiveCfg = Debug|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|x64.Build.0 = Debug|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{72FC9906-07F4-4911-8D6B-F9814974BB37}.Debug|x86.Build.0 = Debug|Any CPU
```
TomerAberbach: Do we know why these get added in addition to "Any CPU"? It seems like the latter would make the former redundant

src/Anthropic/Shims.cs
```diff
@@ -38,4 +39,20 @@ internal sealed class MaybeNullWhenAttribute(bool returnValue) : Attribute;
...
+        {
+            if (string.IsNullOrWhiteSpace(value))
+            {
+                throw new System.ArgumentException(
+                    $"The parameter '{fieldName}' should be not null or an empty string but was either."
+                );
+            }
+        }
+    }
+}
```
TomerAberbach: This will probably become unnecessary after my other comments

src/Anthropic.Tests/Anthropic.Tests.csproj
```diff
@@ -25,6 +25,9 @@
     <PackageReference Include="xunit.runner.visualstudio" Version="2.5.3" />
     <ProjectReference Include="..\Anthropic\Anthropic.csproj" />
     <ProjectReference Include="..\Anthropic.Foundry\Anthropic.Foundry.csproj" />
+    <ProjectReference Include="..\Anthropic.Bedrock\Anthropic.Bedrock.csproj" />
+  </ItemGroup>
+  <ItemGroup>
```
TomerAberbach: Why split the item groups?

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+    private AnthropicBedrockClient(
+        IAnthropicBedrockCredentials bedrockCredentials,
+        ClientOptions clientOptions
+    )
+        : base(clientOptions)
+    {
+        _bedrockCredentials = bedrockCredentials;
+    }
+
+    // <inheritdoc />
```
TomerAberbach:
```suggestion
    /// <inheritdoc/>
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+    public override IAnthropicClient WithOptions(Func<ClientOptions, ClientOptions> modifier)
+    {
+        return new AnthropicBedrockClient(_bedrockCredentials, modifier(this._options));
+    }
+
+    // <inheritdoc />
```
TomerAberbach:
```suggestion
    /// <inheritdoc />
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+            && requestMessage.RequestUri.Segments[3].Trim('/') is "batches" or "count_tokens"
+        )
+        {
+            throw new AnthropicInvalidDataException(
+                $"The requested endpoint '{requestMessage.RequestUri.Segments[3].Trim('/')}' is not yet supported."
+            );
+        }
+    }
+
+    // <inheritdoc />
```
TomerAberbach:
```suggestion
    /// <inheritdoc/>
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+        {
+            return;
+        }
+
+        if (
+            string.Equals(
+                httpResponseMessage.Content.Headers.ContentType?.MediaType,
+                ContentTypeAwsEventStream,
+                StringComparison.CurrentCultureIgnoreCase
+            ) != true
```
TomerAberbach:
```suggestion
            !string.Equals(
                httpResponseMessage.Content.Headers.ContentType?.MediaType,
                ContentTypeAwsEventStream,
                StringComparison.CurrentCultureIgnoreCase
            )
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+        //
+        // Print the SSE (with a blank line after) to the piped output stream to complete the
+        // translation process.
+
+        var originalStream = await httpResponseMessage
+#if NET
+            .Content.ReadAsStreamAsync(cancellationToken)
+#else
+            .Content.ReadAsStreamAsync()
+#endif
```
TomerAberbach: nit: throughout, can we do the `#if NET` wrapping around just the cancellation token param? That's what we do everywhere else

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,272 @@
...
+#if NET
+        httpResponseMessage.Content.Headers.ContentType = new(
+            ContentTypeSseStreamMediaType,
+            "utf-8"
+        );
+#else
+        httpResponseMessage.Content.Headers.ContentType = new(
+            $"{ContentTypeSseStreamMediaType}; charset=utf-8"
+        );
+#endif
```
TomerAberbach: nit: do the wrapping around the params only

src/Anthropic.Bedrock/AnthropicBedrockPrivateKeyCredentials.cs
```diff
@@ -0,0 +1,81 @@
...
+/// <summary>
+/// Provides AWS private key-based credentials for authenticating requests to Amazon Bedrock's Anthropic models.
+/// </summary>
+/// <remarks>
+/// This class implements the AWS Signature Version 4 signing process to authenticate HTTP requests
+/// to Amazon Bedrock using AWS access key and secret key credentials. The credentials are applied
+/// to each request by adding the necessary AWS authentication headers including the authorization
+/// signature, date, and content hash.
+/// </remarks>
+public class AnthropicBedrockPrivateKeyCredentials : IAnthropicBedrockCredentials
```
TomerAberbach: Probably should be `sealed`

src/Anthropic.Bedrock/AnthropicBedrockPrivateKeyCredentials.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// <summary>
+    /// Gets the ApiSecret.
+    /// </summary>
+    public string ApiSecret { get; private init; }
+
+    /// <summary>
+    /// Gets the ApiKey.
+    /// </summary>
+    public string ApiAccessKey { get; private init; }
```
TomerAberbach: Similar comment as other file around construction

src/Anthropic.Bedrock/AnthropicBedrockPrivateKeyCredentials.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    // <inheritdoc />
```
TomerAberbach:
```suggestion
    /// <inheritdoc/>
```

Gotta do this everywhere

TomerAberbach (commented):
I've not looked at everything, but sending what I currently have

src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
JPVenson: They cannot be null and required still lets you set them to null

TomerAberbach: I don't follow. If we do `public string required`, then the user cannot set to null unless they ignore the type system? If they ignore the type system and their code crashes, then I don't care about that case

JPVenson: eh fair point

src/Anthropic.Tests/Anthropic.Tests.csproj
JPVenson: Its purely organisational for better manual editing

Anthropic.sln
JPVenson: No not really, its what the dotnet scaffolding tool adds i never really question it

Issue #53, opened 2025-12-01
Error: tool_use ids must be unique when doing parallel requests


bryansusu:
Hi,

I'm getting the following error when using the official Anthropic SDK together with the Microsoft Agent Framework:

```json
{"type":"error","error":{"type":"invalid_request_error","message":"messages.15.content.2: tool_use ids must be unique"}}
```

This only seems to happen when the agent triggers multiple tool calls in parallel. I don't manually set tool IDs.

Expected: the SDK should either accept parallel tool calls or auto-ensure unique IDs instead of failing.

Could you clarify how tool IDs should be handled, and whether this is a bug or expected behavior?

Thanks!

1yefuwang1:
I'm facing the same issue using the Microsoft.Extensions.AI.IChatClient interface with streaming enabled if the model returns multiple tool calls.

I did some debugging:
In streaming mode, tool1 is returned first then tool2, then tool3.
The final tool list of this SDK is `[tool1, tool1, tool2, tool1, tool2, tool3]`, where tool1 is duplicated 3 times and tool2 twice.

So, I think the bug is in the IChatClient adaptor which treats the whole streaming response(response1: [too1], response2[tool1, too2], resopnse3[tool1, tool2,tool3]) so far as a delta update, leading to duplicated tool_use messages

TomerAberbach:
@stephentoub @rogerbarreto is this perhaps a bug in the custom code?

stephentoub:
> [@stephentoub](https://github.com/stephentoub) [@rogerbarreto](https://github.com/rogerbarreto) is this perhaps a bug in the custom code?

Likely. I will look.

stephentoub:
@bryansusu and/or @1yefuwang1, can you please share a minimal repro demonstrating the issue?

stephentoub:
Nevermind, I was able to repro. I see the problem. Easy fix.

stephentoub:
@TomerAberbach, this should be fixed by https://github.com/anthropics/anthropic-sdk-csharp/pull/62

bryansusu:
@stephentoub any idea when v12 will be released?

stephentoub:
I don't know. @sd-st? @TomerAberbach ?

sd-st:
I did a v12 release earlier today but it looked like it missed some changes. We're going to look into it a bit tomorrow to see what happened and hopefully get a new release out soon.

TomerAberbach:
[...]

PR #62, opened 2025-12-03
Fix handling of multiple streaming tool calls

stephentoub:
The implementation was missing a clear on the collection of streaming data. As such, it would yield each tool already yielded every time it yielded another tool.

Fixes https://github.com/anthropics/anthropic-sdk-csharp/issues/53


stephentoub:
cc: @TomerAberbach, @sd-st

sd-st (approved):

bryansusu:
Great job guys. Looking forward to test the fix.
When is this expected to be released?

PR #63, opened 2025-12-04
Add beta headers automatically for HostedCodeInterpreter/McpServerTool

stephentoub:
cc: @sd-st, @pederhp

Per discussion at https://github.com/anthropics/anthropic-sdk-csharp/pull/58#discussion_r2586369119.


PederHP (approved):

sd-st (approved):

sd-st:
Looks like this needs a format (can do with `./scripts/format`) and then it should be good to merge. Probably also worth doing a `./scripts/lint` to make sure everything is happy

PR #65, opened 2025-12-08
Fix/server tool use and union deserialization

bQvle:
fix: type-checked union deserialization and optional field corrections

Fixes server-side tool deserialization by:
- Checking 'type' field before deserializing unions (eliminates silent exceptions)
- Making incorrectly-required fields optional (Caller, WebFetchRequests, WebSearchRequests)
- Routing directly to correct deserializer instead of try-catch approach (which caused actual bugs to get silently bypassed)
- Throwing exception for unhandled types

This enables server_tool_use / web_search and web_fetch tools to work correctly.


src/Anthropic/Models/Beta/Messages/BetaToolUnion.cs
```diff
@@ -1004,291 +1004,230 @@ JsonSerializerOptions options
...
+            default:
+                // Unknown tool type - throw exception with details so we can add support
+                var jsonStr = json.GetRawText();
+                throw new AnthropicInvalidDataException(
+                    $"Unknown tool type '{typeStr}'. Please add support for this tool type. JSON: {jsonStr}"
+                );
         }
 
+        // If we get here, deserialization failed - return raw JSON as fallback
         return new(json);
```
bQvle, line 1231: Don't know what is best, to return the raw json, or throw in default case. I don't know in what case it is useful to just return raw json. might be my inexperience.

sd-st:
Hey @bQvle, thanks for this PR! I think we should update the OpenAPI spec on our end to reflect these changes which will then result in the correct code getting generated. cc @felixfbecker

sd-st:
Closing as this will be an OpenAPI spec change

PR #66, opened 2025-12-08
Use Random.Shared on .NET Core

stephentoub:


sd-st:
Sorry, can you explain why we want this?

stephentoub:
`Random.Shared` on .NET Core provides a thread-safe Random singleton. The current code is trying to replicate a similar thing by using a ThreadLocal and then a separate Random instance per thread. That's fine for netstandard2.0 where Random.Shared doesn't exist, but where it does exist, it's better to just use the built-in solution for this exact need.

sd-st:
FYI we are backporting this to codegen so this should be implemented soon. Thanks for your PR!

stephentoub:
https://github.com/anthropics/anthropic-sdk-csharp/commit/f87fa6582c33d0fc124c3c08db9911b2ea46b0b0

PR #67, opened 2025-12-08
Avoid unnecessary CancellationTokenSource

stephentoub:


TomerAberbach:
Is there any behavior difference with this? Or is it just a more terse way of doing the same thing?

stephentoub:
> Is there any behavior difference with this? Or is it just a more terse way of doing the same thing?

Same behavior but cheaper and less code.

sd-st:
This is something I think we'd want to do as part of codegen, otherwise there's an opportunity for merge conflicts with generated code in the future. I'm going to close this, though if it's important to you that we implement it feel free to open an issue.

stephentoub:
@sd-st, what is the plan for moving into the repo whatever files are part of "codegen"? It's not much an open source project if PRs aren't permitted to touch the bulk of the implementation.

sd-st:
The codegen logic is currently closed source although we do have plans to open source it. In general we're happy to backport improvements into codegen ourselves, although in this case it's hard to justify prioritizing as the benefits seem quite minor (please correct me if I'm wrong)

stephentoub:
Thanks. It's not about this particular change being super meaningful or not, but rather that it'd be nice if the repo were accepting of PRs, even for minor improvements, but if the vast majority of the code is codegen'd and the codegen support is closed source, the repo isn't really open.

Issue #40, opened 2025-11-26
Add Strong Name signing to the Anthropic libraries.


rogerbarreto:
Adding a Strong Name to Anthropic will allow it to be used as reference into strong named assemblies like Microsoft Agent Framework and many other .Net Standard projects.

Since the "Anthropic" package does not have a strong name, Microsoft Agent framework and many other customers targeting .Net Standard, won't be able to use the package.

Those strong name issues do not apply to .Net Core though.

Potential errors users of `Microsoft.Agents.AI.Anthropic` will face if using non-strong-name signed from Anthropic

### Checking script
```powershell
$package='Anthropic';$ver='10.4.0';$url="https://www.nuget.org/api/v2/package/$package/$ver"
Invoke-WebRequest $url -OutFile "$package.$ver.nupkg"
Expand-Archive "$package.$ver.nupkg" -DestinationPath $package -Force
Get-ChildItem $package -Recurse -Filter *.dll | Where-Object {$_.FullName -like '*\lib\*'} | ForEach-Object {
    & sn -vf $_.FullName
}
Remove-Item "$package.$ver.nupkg" -Force
```

### Current results
```
Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

D:\repo\community\anthropic-sdk-csharp\temp\Anthropic\lib\net8.0\Anthropic.dll does not represent a strongly named assembly

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

D:\repo\community\anthropic-sdk-csharp\temp\Anthropic\lib\net9.0\Anthropic.dll does not represent a strongly named assembly

Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

D:\repo\community\anthropic-sdk-csharp\temp\Anthropic\lib\netstandard2.0\Anthropic.dll does not represent a strongly named assembly
```

### Expected result
```
Microsoft (R) .NET Framework Strong Name Utility  Version 4.0.30319.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Assembly 'D:\repo\community\anthropic-sdk-csharp\temp\Anthropic\lib\netstandard2.0\Anthropic.dll' is valid
```

[...]

PR #68, opened 2025-12-09
Enable strong-name signing

stephentoub:
Enables other strong-name-signed netstandard2.0 assemblies to depend on Anthropic{.Foundry}.

Fixes https://github.com/anthropics/anthropic-sdk-csharp/issues/40


sd-st:
We do want to implement strong-name signing, but internally we've discussed it and I don't believe we want to put the private key in the repo. We were planning to expose it as an environment variable within GitHub Actions, which we would then use for signing.

ericstj:
We do recommend that folks make their key available for anyone to use.  https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/strong-naming

In this case the Open.snk is the same that we make publicly available in https://github.com/dotnet/arcade/blob/main/src/Microsoft.DotNet.Arcade.Sdk/tools/snk/Open.snk.  There's no good reason to keep it secret.

rogerbarreto:
Initially I had second thoughts about exposing the Private/Public key as part of the `snk` file. But seems evident for the sake of `Strong Name` signing this is not an issue.

From my own experience regarding `public/private` key generation, normally the private is held secretly using the delayed signing approach, but OSS seems to not follow this case and I think is ok to moving forward as the PR suggest.

sd-st:
Cool, makes sense. In that case I'm going to backport this to codegen.

HaripriyaMehta:
Thank you for working on this! I'm interested in strong name-signed support as well. Is there an ETA for when a NuGet package with these changes might be available?

stephentoub:
> Cool, makes sense. In that case I'm going to backport this to codegen.

So I should close this?

sd-st:
> So I should close this?

Yes please! I'll let you know when this is implemented.

> Thank you for working on this! I'm interested in strong name-signed support as well. Is there an ETA for when a NuGet package with these changes might be available?

I don't have a precise ETA but we're planning on working on this fairly soon. Apologies for missing this; we just got back from holiday break

PR #71, opened 2025-12-10
Fix build warnings in next

stephentoub:
There are currently 62 build warnings. They're driving me a bit nuts :) Fixing them all, and turning on warnings as errors (I've added a Directory.Build.props... more centralized settings should subsequently be moved from the csproj into this).


sd-st:
Hey @stephentoub! Thanks for the PR as always. We're actually planning to tackle a bunch of this in codegen. In general it may be worth filing an issue before opening PRs, because generally we'll backport anything we can to codegen on our end.

PR #72, opened 2025-12-12
MicrosoftExtensions: Normalize nullable union tool types

PederHP:
See: https://github.com/modelcontextprotocol/csharp-sdk/issues/1091

Tools with `["integer", "null"]` discriminated union style types to provide additional information on default when non provided for non-required properties would cause the following error:

`tools.0.custom.input_schema: JSON schema is invalid. It must match JSON Schema draft 2020-12 (https://json-schema.org/draft/2020-12). Learn more about tool use at https://docs.claude.com/en/docs/tool-use.`

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
+++ b/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
@@ -2304,4 +2304,174 @@ public class AnthropicClientBetaExtensionsTests : AnthropicClientExtensionsTests
         Assert.Equal("application/pdf", files[1].MediaType);
         Assert.Equal(2000, files[1].SizeInBytes);
     }
+
+    [Fact]
+    public async Task GetResponseAsync_WithNullableUnionType_TransformsToSimpleType()
+    {
+        // The C# MCP SDK generates schemas with "type": ["integer", "null"], "default": null
+        // for optional nullable parameters. This test verifies the transformation removes
+        // the null from the union type and the default: null for non-required properties.
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Call tool with optional param"
+                    }]
+                }],
+                "tools": [{
+                    "name": "tool_with_optional",
+                    "description": "A tool with an optional nullable parameter",
+                    "input_schema": {
+                        "type": "object",
+                        "properties": {
+                            "required_param": {
+                                "type": "string"
+                            },
+                            "optional_number": {
+                                "type": "integer",
+                                "description": "{default: null}"
+                            }
+                        },
+                        "required": ["required_param"],
+                        "additionalProperties": false
+                    }
+                }]
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_nullable_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Tool ready"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 30,
+                    "output_tokens": 5
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        // Create a function with an optional nullable parameter - the schema will have
+        // "type": ["integer", "null"], "default": null which should be transformed
+        var functionWithOptional = AIFunctionFactory.Create(
+            (string required_param, int? optional_number = null) => "result",
+            new AIFunctionFactoryOptions
+            {
+                Name = "tool_with_optional",
+                Description = "A tool with an optional nullable parameter",
+            }
+        );
+
+        ChatOptions options = new() { Tools = [functionWithOptional] };
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Call tool with optional param",
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithRequiredNullableUnionType_PreservesUnionType()
+    {
+        // When a property IS in the required array but has a nullable union type,
+        // we should NOT transform it - let the API fail with a meaningful error
+        // rather than silently misrepresenting the schema.
+        // Using BetaTool.AsAITool() bypasses the schema transformation, so this test
+        // verifies the raw tool passes through unchanged.
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Call tool with required nullable"
+                    }]
+                }],
+                "tools": [{
+                    "name": "tool_with_required_nullable",
+                    "description": "A tool with a required nullable parameter",
+                    "input_schema": {
+                        "nullable_number": {
+                            "type": ["integer", "null"],
+                            "description": "A required but nullable number"
+                        },
+                        "type": "object",
+                        "required": ["nullable_number"]
+                    }
+                }]
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_required_nullable_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Tool ready"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 30,
+                    "output_tokens": 5
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        // Create a BetaTool with a required nullable parameter using raw schema
+        // This simulates a schema that came from elsewhere (not C# MCP SDK pattern)
+#pragma warning disable CA1861 // Prefer 'static readonly' fields over constant array arguments - test method only runs once
+        BetaToolUnion rawTool = new BetaTool
+        {
+            Name = "tool_with_required_nullable",
+            Description = "A tool with a required nullable parameter",
+            InputSchema = new InputSchema(
+                new Dictionary<string, JsonElement>
+                {
+                    ["nullable_number"] = JsonSerializer.SerializeToElement(
+                        new
+                        {
+                            type = new[] { "integer", "null" },
+                            description = "A required but nullable number",
+                        }
+                    ),
+                }
+            )
+            {
+                Required = ["nullable_number"],
+            },
+        };
+#pragma warning restore CA1861
+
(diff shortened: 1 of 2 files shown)


PederHP:
@stephentoub Any comments on this one? I made it to unbreak use of MCP Server tools which have discriminated unions from default parameters (which are valid JSON schema but not supported by the Anthropic API). I assume the code owners will lean on your review of this as it relates only to MEAI.

It would be really nice to get this in as it as otherwise the risk of MCP tools breaking the client is quite high with the beta client - requiring the use of a forked SDK. If the stricter schema validation of the beta client makes it to the main client, I think this will affect almost anyone using MCP with the SDK and extensions.

stephentoub:
@eiriktsarpalis, could you take a look at this one?

src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
```diff
@@ -2214,4 +2214,168 @@ public async Task GetResponseAsync_WithHostedToolsAndExistingBetas_PreservesAndD
         Assert.Contains("code-execution-2025-08-25", capturedBetaHeaders);
         Assert.Contains("mcp-client-2025-11-20", capturedBetaHeaders);
     }
+
+    [Fact]
+    public async Task GetResponseAsync_WithNullableUnionType_TransformsToSimpleType()
+    {
+        // The C# MCP SDK generates schemas with "type": ["integer", "null"], "default": null
```
eiriktsarpalis, line 2311: I'm not sure I entirely follow the motivation for this change. As discussed in https://github.com/modelcontextprotocol/csharp-sdk/issues/1091 the schema can be influenced by making corresponding changes in the method signature, namely:

1. The `null` component from the `type` can be removed by making the parameter type non-nullable and
2. The `default` keyword can be removed by removing the default value from the parameter.

Is there something that prevents users from making these types of changes? Or something different?

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -160,6 +160,87 @@ private sealed class AnthropicChatClient(
                         {
                             _ = schemaObj.Remove(propName);
                         }
+
+                        // Handle nullable union types (e.g., "type": ["integer", "null"]) for
+                        // non-required properties. The C# MCP SDK generates these for optional
+                        // parameters, but structured outputs doesn't support union types.
```
eiriktsarpalis: Is the form `"type" : ["integer","null"]` considered a union type? I would have guessed such a terminology would be reserved for schemas using `"anyOf"` or `"oneOf"`. Is the trivial schema `{}` (which we generate for things like `object` or `JsonElement`) similarly considered a union type in that case, and would we need to similarly restrict those should they turn up?

eiriktsarpalis (commented):
The fact that this is constrained to a particular chat client makes me feel less concerned, however I do think that a nullable/optional input parameter encodes important aspects of an AIFunction's contract, so silently erasing those is likely going to result in unexpected behavior.

I would much rather have the chat client fail with an explicit "not supported" error message and give the user to opportunity to model their functions in a way that remains compatible with the LLM in question.

src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
stephentoub: > Is there something that prevents users from making these types of changes? Or something different?

This is a client, handed arbitrary tools, with zero control over the schema used by those tools when they come from another party's mcp server.

stephentoub:
> I would much rather have the chat client fail with an explicit "not supported" error message and give the user to opportunity to model their functions in a way that remains compatible with the LLM in question.

Then why do we expose transforms at all? The fact is that clients are talking to any number of mcp servers they don't own, all that have varying forms of schema employed as the mcp spec says nothing about the format, and talking to llms that each have their own peculiarities. Failing is a bad option in my opinion, as it completely breaks being able to use a valid mcp server with a valid llm.

eiriktsarpalis:
> > I would much rather have the chat client fail with an explicit "not supported" error message and give the user to opportunity to model their functions in a way that remains compatible with the LLM in question.
> 
> Then why do we expose transforms at all?

Marking all inputs as required or transforming `"type" : "null"` to `"nullable" : "true"` is one thing, silently restricting the function inputs is another. The point of the transform is to do just that, transform the advertised schema in a way that remains compatible with the underlying function contract. I could write a transform that collapses all schemas to accepting just one integer, but that wouldn't make it valid transform.

stephentoub:
> The point of the transform is to do just that, transform the advertised schema in a way that remains compatible with the underlying function contract.

Reducing the allowed inputs to a subset is still compatible. If the other excluded inputs are erroneous from the llms perspective, then it wouldn't generate those, anyway.

If there's a better way to express the transformation, great, let's do that. If there isn't, I fail to see the problem here.

eiriktsarpalis:
> Reducing the allowed inputs to a subset is still compatible. If the other excluded inputs are erroneous from the llms perspective, then it wouldn't generate those, anyway.

I'm not questioning the soundness of it, however my point is that it could be limiting the intended functionality of the passed function. Consider the following hypothetical method, which would lose the ability to enumerate all documents:

```C#
// If no owner id is specified, all documents are returned
IEnumerable<Document> GetDocuments(int? ownerId = null);
```

stephentoub:
>  Consider the following hypothetical method, which would lose the ability to enumerate all documents:

If the LLM doesn't support producing null for the input, the transform isn't losing anything.

If the LLM does support producing null for the input, and there's some way to express in the schema to the LLM that the parameter may be null, then the transform should produce the right schema for it.

eiriktsarpalis:
> > Consider the following hypothetical method, which would lose the ability to enumerate all documents:
> 
> If the LLM doesn't support producing null for the input, the transform isn't losing anything.

Yes? I think the point is that the user _wants_ the LLM to possibly provide null and we're silently pushing that part of the contract under the rug. We should instead provide feedback prompting the user to remodel their function into something compatible with the constraints of the current LLM.

stephentoub:
> Yes? I think the point is that the user wants the LLM to possibly provide null and we're silently pushing that part of the contract under the rug.

No, we're not. We have no control over what an LLM does. An LLM that's asked for an integer might only be capable of producing the interger 42, and that's fine. The client is choosing to use that LLM. If a function says it can accept null and the LLM isn't capable of producing null, the LLM just won't produce null; that's not an error, that's implicit to the LLM being used, just as is any other aspect of how good or bad the LLM is at particular workloads. It is not our place to error out if the LLM doesn't meet a particular capability threshold.

PederHP:
I think we should be able to have a client built using the Anthropic SDK and the IChatClient extension connect to arbitrary MCP servers without getting runtime errors due to the API not accepting discriminated unions.

LLMs consider unspecified parameters as something different than null. The LLM is perfectly capable of not passing an optional int parameter.

The user can't remodel functions because they're neither the client or server developer.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: I will have to test if {} causes an error. Note that the API just rejects the request without more information than shown in the PR description. Calling it a (discriminated) union type is on me. If there’s a more accurate description I am happy to change the comment.

But the API definitely rejects requests if this pattern is present.

sd-st:
Hey sorry, just looking at this now. Looks like this could use a rebase which is messing up the diff so I'm not exactly sure what the code changes are here but I read through the discussion.

If I'm understanding things correctly it seems like transforming from `type: ["null", "integer"]` to a nullable integer is fine? They're functionally equivalent so regardless of if the LLM supports nulls it shouldn't make a difference. Maybe I'm misunderstanding something?

stephentoub:
> Looks like this could use a rebase which is messing up the diff

As an aside, @sd-st, rewriting history in the next branch seems to be pretty common. Any chance we could stop doing that? :smile:

sd-st:
> > Looks like this could use a rebase which is messing up the diff
> 
> As an aside, @sd-st, rewriting history in the next branch seems to be pretty common. Any chance we could stop doing that? 😄

You're right, this is definitely annoying. The reason we do this is to keep `next` strictly ahead of `main`. Ideally we could merge `next` into `main` in a way that doesn't require a rewrite, but in practice we'll sometimes do little things like clean up commit messages. Unfortunately there's no good way for us to stop doing this without changing our strategy around how releases work. This is something we have on our roadmap although I unfortunately don't have an ETA.

stephentoub:
Ok, thanks

PederHP:
I will get a rebase done.

PederHP:
> If I'm understanding things correctly it seems like transforming from `type: ["null", "integer"]` to a nullable integer is fine? They're functionally equivalent so regardless of if the LLM supports nulls it shouldn't make a difference. Maybe I'm misunderstanding something? 

I agree. A server which maps nullable integer to this type shouldn't mind an optional integer on the request side. A tool call without the parameter will have it mapped to null before the mcp tool invocation anyway. So functionally equivalent.

And it means I won't need a patched SDK to allow user configured MCP connections (which may be to servers returning this incompatible schema).

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -197,6 +197,87 @@ is string version
                         {
                             _ = schemaObj.Remove(propName);
                         }
+
+                        // Handle nullable union types (e.g., "type": ["integer", "null"]) for
+                        // non-required properties. The C# MCP SDK generates these for optional
+                        // parameters, but structured outputs doesn't support union types.
+                        // Only transform non-required properties; required + nullable is an
+                        // incompatible schema that should fail at the API level.
```
sd-st: Hmm conceptually I don't agree with this comment. There's a difference between `{}`, `{ foo: null }`, and `{ foo: "bar" }`. A required + nullable property is one that must be specified (but may be specified to be null). In the context of an LLM they are pretty much interchangeable but I wouldn't say it's invalid

May be worth updating the comment and possibly also changing the logic

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1521,6 +1605,77 @@ out Uri? url
             return annotation;
         }
 
+        /// <summary>
+        /// Transforms a property schema with nullable union type (e.g., "type": ["integer", "null"])
+        /// to a simple type (e.g., "type": "integer") and removes "default": null.
+        /// This handles schemas generated by the C# MCP SDK for optional nullable parameters.
+        /// </summary>
+        private static JsonElement TransformNullableUnionType(JsonElement propertySchema)
```
sd-st: This seems quite similar to the code up above. Can they be combined/deduped?

sd-st:
@PederHP Thanks for rebasing! Just flagging I have a few unaddressed comments still

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: I agree. But when I tested it the API failed when sending required + nullable. I will re-test and change the comment to make it more clear that the reason for this logic is that the API fails on it, not that it is conceptually invalid.

PederHP:
@sd-st The original issue can no longer be reproduced. I think the validator / json parser in the API now accepts this schema, so no need for the change at all! That's perfect. Dropping the PR.

PR #73, opened 2025-12-14
MicrosoftExtensions: Idiomatic CacheControl Support

PederHP:
For agentic loops the cost difference between caching and not caching is massive, and it's really awkward to use caching with the extensions. And not really possible at all at the system-level (because of how it converts without using the raw representation factory). Using raw representation factory also means one has to re-implement the mapping from MEAI to Anthropic content types. Instead this PR adds a new extension method that uses AdditionalProperties to store the CacheControl and then uses it while mapping in the existing extension code.


PederHP:
Let me know if you want a separate extension for beta and non-beta cache control (currently the outer one is shared). I also added a new file. I can fold it into the extensions file, but it's getting massive, so I think if it's not ok to have more than one top-level MEAI related file - then maybe there should be a folder? Otherwise the file is just going to keep growing and get more and more unwieldy.

sd-st:
Thanks!

PR #75, opened 2025-12-15
Update to M.E.AI.Abstractions 10.1.1

stephentoub:


src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -1295,7 +1295,7 @@ public async Task GetResponseAsync_WithCacheCreationTokens()
         Assert.Equal(185, response.Usage.TotalTokenCount);
         Assert.NotNull(response.Usage.AdditionalCounts);
         Assert.Equal(50L, response.Usage.AdditionalCounts["CacheCreationInputTokens"]);
-        Assert.Equal(25L, response.Usage.AdditionalCounts["CacheReadInputTokens"]);
+        Assert.Equal(25L, response.Usage.CachedInputTokenCount);
```
PederHP, line 1298: Is this the same metric? I think `CachedInputTokenCount` sounds like Cache Write and not Cache Read (hit)? Or at least it's somewhat ambiguous. 

Probably a bit late now that it's already named in MEAI - if this is actually the name of the cache hit token count. But I think having Read or Hit in the name would make it much clearer that this is not a cache write.

Edit: Ah - I see now, it's lifted from the OpenAI naming conventions. I think it's a shame that it's not named more clearly, but probably too late now. It's not unclear when looking at the full UsageDetails datastructure, but in isolation it would be much clearer if it included the Read verb in the property name.

sd-st (approved):

PR #77, opened 2025-12-18
 Add Bedrock as provider for SDK

JPVenson:
Adds Bedrock implementation for usage with api.

Notes:
Only the Model invoke and streaming endpoint is currently supported

this is the successor pr for #61


sd-st:
FYI this has a failing lint and build

JPVenson:
@sd-st fixed

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,183 @@
...
+    )
+    {
+        var (data, success) = await ReadStreamMessage(source, cancellationToken)
+            .ConfigureAwait(false);
+        if (!success)
+        {
+            return false;
+        }
+        await target
+            .WriteAsync(Encoding.UTF8.GetBytes(data!), cancellationToken)
```
sd-st: NIT: Wonder if we should return false if data is null instead of asserting it's not null here

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,183 @@
...
+        events come in the form of the event stream.
+        https://docs.aws.amazon.com/lexv2/latest/dg/event-stream-encoding.html
+        |                Prelude                  |             |            Data                 |
+        | Total Byte Length | Headers Byte Length | Prelude CRC | Headers | Payload | Message CRC |
+        | 4 byte            | 4byte               | 4byte       |                                 |
+        */
+
+        Span<byte> preamble = stackalloc byte[12];
+        var preambleRead = source.Read(preamble);
+        if (preambleRead != 12 || !Crc32ChecksumValidation(preamble[..8], preamble[8..]))
```
sd-st: Hmm I don't think I understand how the logic here works. Surely it's possible that `source.Read` happens right while the internal buffer is getting filled (e.g. from network I/O), in which case we could end up throwing those bytes away and throwing an exception? 

From https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-10.0:
> The total number of bytes read into the buffer. This can be less than the size of the buffer if that many bytes are not currently available, or zero (0) if the buffer's length is zero or the end of the stream has been reached.

I'm also getting a little concerned with how low level we're getting here; it looks like e.g. for Java we just use the AWS library: https://github.com/anthropics/anthropic-sdk-java/blob/main/anthropic-java-bedrock/src/main/kotlin/com/anthropic/bedrock/backends/BedrockBackend.kt
Can we do the same thing here?

src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -0,0 +1,26 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
sd-st, line 1: Not sure where to comment this but I think an example in the `examples` folder would be good

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,280 @@
...
+            contentStream.Seek(0, SeekOrigin.Begin);
+            requestMessage.Headers.TryAddWithoutValidation(
+                "content-length",
+                contentStream.Length.ToString()
+            );
+            var strUri =
+                $"{requestMessage.RequestUri!.Scheme}://{requestMessage.RequestUri.Host}/model/{modelValue}/{(parsedStreamValue ? "invoke-with-response-stream" : "invoke")}";
+
+#if NET6_0_OR_GREATER
+            // The UriCreationOptions and DangerousDisablePathAndQueryCanonicalization were added in .NET 6 and allows
```
sd-st, line 148: Curious what happens pre-NET6? Does it canonicalize the URL or no? Just wanna make sure we support all versions properly

sd-st (commented):
I did an initial skim; just wondering if we can use more of the AWS library instead of rolling out own

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: The base implementation here is the httpbufferedstream which will wait until the requested chunck is available or return whats available when the remote has closed the connection. Thats how the scrapped async variant i proposed initially would have also worked.

The AWS lib for c# does have an implementation, but its private so we cannot access it at all and would need to fully use their API instead of our own which would be a massive project (different models). The same goes for the authentication part, the AWS lib does have that too but not accessable for us without reflection shananigans.

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: It does not. This is a long standing issue with AWS url format and is discussed quite a bit on their api too. From which i took this snipped.

TL;DR: thats the way AWS does it

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: I see, in that case do we need to make the `Stream` parameter type more constrained?

JPVenson: i would not say so no.
That class isn't _really_ supposed to be called by anyone else and its the contract behind its usage which is pretty standard for streams

src/Anthropic.Tests/AnthropicTestClients.cs
```diff
@@ -3,6 +3,7 @@
 using System.Linq;
```
sd-st, line 3: Curious if we should have some more bedrock-specific testing? Like that the requests it makes are what we expect and have the right content etc.

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,183 @@
...
+/// the AWS event stream encoding specification.
+/// </summary>
+/// <remarks>
+/// Reference: https://docs.aws.amazon.com/lexv2/latest/dg/event-stream-encoding.html
+/// </remarks>
+internal static class SseEventHelpers
+{
+    private static readonly JsonSerializerOptions? _jsonOptions = new() { WriteIndented = false };
+
+    public static async Task<bool> SyncStreamMessage(
```
sd-st: This method doesn't actually seem to be used anywhere. Do we need it?

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: I think I need to understand this a bit more. It looks like this stream comes from `ReadAsStreamAsync()` which returns a normal `Stream`. The `Stream.Read` explicitly calls out that reads can be partial:

https://learn.microsoft.com/en-us/dotnet/api/system.io.stream.read?view=net-10.0

> Returns: The total number of bytes read into the buffer. This can be less than the size of the buffer if that many bytes are not currently available, or zero (0) if the buffer's length is zero or the end of the stream has been reached.

Basically I'd like to know why in this case we don't have to worry about partial reads. I want to make sure we're not violating any contracts

JPVenson: `Stream` is an abstract class and needs to be implemented for your usage. That class itself does not define any real logic on data manipulation.

Think of it as a fancy interface but with pre-defined methods in some places. The implementation for the GetStreamAsync from the HttpClient blocks until the requested data is read

GetStreamAsync
> This operation doesn't block. The returned [Task<Stream>](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task-1?view=net-10.0) object will complete after the response headers are read. This method does not read nor buffer the response body.

The data is then read on demand with the `Read` methods. A 0 Read means that no more data is available and the operation is done

sd-st: Sorry, heading into a meeting but I'm going to reread this after and make sure I understand

JPVenson: https://github.com/dotnet/runtime/issues/31316

This gives a pretty good overview about the topic of buffered vs non-buffered usage of the HttpContent behavior. Its really badly documented on MSDN so i totally understand your confusion. I only speak from experience working with it.

sd-st: I don't that thread really clears things up for me. Like regardless of whether it's buffered or not it seems totally possible for `Read` to only read and return a subset of bytes like it's documented as doing. And even if it doesn't now, it seems like in the future this could change and it would be a non-breaking change, since it's not a contract anywhere.

@stephentoub I see you commented on the linked thread, wondering if you have any thoughts here?

stephentoub:
I'm curious as to the motivation for this PR, namely when I would use this library vs Amazon's Bedrock runtime library (which also provides an IChatClient).

src/Anthropic.Bedrock/SseEventHelpers.cs
stephentoub: Anything working in terms of the abstract Stream should be written to expect that Read may return fewer than the requested number of bytes.

> The implementation for the GetStreamAsync from the HttpClient blocks until the requested data is read

HttpClient.GetStreamAsync returns a Task that will complete once the headers have been read, giving back a Stream that can then be used to consume the body. That stream is _not_ fully buffered, meaning it doesn't buffer the full response and then hand out parts of that full response. Reads on it are in no way guaranteed to return as much data as was asked for, and frequently won't.

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,183 @@
...
+
+/// <summary>
+/// Provides helper methods for processing Server-Sent Events (SSE) from AWS Bedrock event streams.
+/// Handles parsing, validation, and synchronization of event stream messages encoded according to
+/// the AWS event stream encoding specification.
+/// </summary>
+/// <remarks>
+/// Reference: https://docs.aws.amazon.com/lexv2/latest/dg/event-stream-encoding.html
+/// </remarks>
+internal static class SseEventHelpers
```
stephentoub: This says "SSE" and "Server-Sent Events", but this looks like it's actually handling Amazon's event stream encoding? What is the relationship?

sd-st:
> I'm curious as to the motivation for this PR, namely when I would use this library vs Amazon's Bedrock runtime library (which also provides an IChatClient).

We want people who are using the official Anthropic SDK to be able to use Bedrock. We're adding support for other third party runners as well

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: i qouted the wrong method sory. We are calling this from the stream returned from the `HttpContent.ReadAsStreamAsync` method not the `HttpClient.GetStreamAsync` method. The former does actually have a different behavior if not buffered. You are technically right as we are only handling the stream here, as a compromise i would just make the class internal so we can keep relying on the implied implementation behavior as expected

stephentoub: > We are calling this from the stream returned from the HttpContent.ReadAsStreamAsync method not the HttpClient.GetStreamAsync method. The former does actually have a different behavior if not buffered.

HttpContent.ReadAsStreamAsync is also not necessarily going to be reading fully buffered data and typically won't be. If the request was made to HttpClient with HttpCompletionOption.ResponseHeadersRead, or if you're using the underlying HttpMessageHandler directly (e.g. with HttpMessageInvoker), then if you do response.Content.ReadAsStreamAsync, it will not fully buffer the response content and reads will very likely come back with less than you ask for.

JPVenson: But it is reading the full buffer size which far exceeds what we need here so i dont really see a problem in this case. If i change this to `ReadExactly` anyway would that be solving this discussion?

stephentoub: > But it is reading the full buffer size which far exceeds what we need here so i dont really see a problem in this case.

Sorry, I'm not following. Are you saying you're passing a large number to read and you only need a small number so it shouldn't be an issue? You could request a million bytes and it'd be valid for read to return you just one. When using Stream.Read{Async}, one always needs to check that the requested number of bytes was actually read, and if you need more to continue, read more, in a loop.

> If i change this to ReadExactly anyway would that be solving this discussion?

ReadExactly does such a loop, so yes, if you know you need 12 bytes, you can ReadExactly 12 bytes, or ReadAtLeast 12 bytes.

(As an aside, this should really be a ReadAsync rather than Read.)

JPVenson: edit: nvm missread the original comment. Doesn't matter to discuss this further anyway, i change it to ReadExactlyAsync and everyones happy

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: Everything from amazon follows that type of encoding which is also the reason it took me so long to find documentation about it^^

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: changed, but i used `ReadExactly` here instead of the Async variaents because they ofc require `Memory<>` which would negate the benefits of the current design

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: nope

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: method removed

src/Anthropic.Bedrock/SseEventHelpers.cs
stephentoub: And they call it SSE?

JPVenson: Yesn't.

Anthropic calls it ServerSendEvents generally which for AWS is using the EventStreamEncoding. Everything in AWS is encoded with that pattern.

So its our helper for ServerSendEvents which decodes the EventStreamEncoding.

src/Anthropic.Tests/AnthropicTestClients.cs
JPVenson: i was thinking about it, but the only valuable testing points are the SSE intigrations with AWS and that would be a lot of testing effort. I am impartial towards it overall though.

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: Can you explain what these benefits are? `ReadExactly` would mean that we'll block while waiting for bytes, which isn't ideal

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: I kind of wonder if we should leave it as `ServerSendEventsHelpers` or something along those lines? I think people generally associate SSE with HTTP SSE

src/Anthropic.Bedrock/SseEventHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+            if (!Crc32ChecksumValidation([.. preamble, .. header, .. messageSpan.Span], messageCrc))
+            {
+                throw new InvalidDataException(
+                    "The calculated crc checksum for the message content does not match the provided value from the server."
+                );
+            }
+            var result = await Parse(new ReadOnlySequence<byte>(messageSpan), cancellationToken)
+                .ConfigureAwait(false);
+            return (result, true);
+        } // do not catch the EndOfStream exception here as its still unexpected to happen here and should bubble up
```
sd-st: NIT but we generally want to make sure our grammar is consistent, especially as this is open source/public facing 
```suggestion
        } // do not catch the EndOfStream exception here as it's still unexpected to happen here and should bubble up
```

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
```diff
@@ -0,0 +1,168 @@
...
+                _jsonOptions,
+                cancellationToken
+            )
+            .ConfigureAwait(false);
+        if (parsedEvent is null)
+        {
+            return null;
+        }
+
+        return $"event:{parsedEvent["type"]}\ndata:{parsedEvent.ToJsonString(_jsonOptions)}\n\n"; // add double linebreaks at the end to force the StreamReader to emit an empty line for parsing.
```
sd-st: NIT: can we put the comment on the previous line so that it's not so long?

sd-st (commented):
Sorry for all the small reviews, haven't had a chance to do a super in-depth one yet

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: the existing memory buffers are allocated on the stack which is far more efficient memory wise but as the async pattern by definition is a break in the flow of execution and will operate possibly over different points in time on different threads the whole benefit of having this small and efficient goes out the window as we would need to allocate on the heap instead. Its not a huge deal i would say as if you really have to wait for 8-24 bytes of the SSE header (calls down below included) i doubt anything would really work in your app anyway xD.

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: i want to say that you have to keep in mind that this is an internal class that the enduser never sees really. But i am really impartial to the naming^^

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: I think the issue is that the 8-24 bytes has an unbounded time component - you might be on a flaky or slow connection, and it's possible the network hangs right as you're reading these bytes. In this case it would be very bad if we halted the entire thread in a non-async-aware way

src/Anthropic.Bedrock/SseEventHelpers.cs
sd-st: This is true, although it's fairly reasonable to expect that users will look at this code e.g. in GitHub itself, and it would be nice if we named this in a straightforward way

sd-st: Maybe `AwsEventStreamHelpers` is better

src/Anthropic.Bedrock/SseEventHelpers.cs
JPVenson: sure refactoring will take bit though as `ReadExactly` and `ReadExactlyAsync` both need to be shimmed to be used in NetStandard2.0 and `ReadAsync` is a lot more involved then `Read`

src/Anthropic.Tests/Services/Beta/Messages/BatchServiceTest.cs
```diff
@@ -10,7 +10,7 @@ namespace Anthropic.Tests.Services.Beta.Messages;
 public class BatchServiceTest
 {
     [Theory(Skip = "prism validates based on the non-beta endpoint")]
-    [AnthropicTestClients]
```
sd-st, line 13: Why do some of these tests work with the other providers but not bedrock?

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+        | Total Byte Length | Headers Byte Length | Prelude CRC | Headers | Payload | Message CRC |
+        | 4 byte            | 4byte               | 4byte       |                                 |
+        */
+
+        Memory<byte> preamble = new byte[12];
+        try
+        {
+            await source.ReadExactlyAsync(preamble, cancellationToken).ConfigureAwait(false);
+        }
+        catch (EndOfStreamException) // exceptions for control flow are bad but its how this works. Only catch it here as at every other occasion it's not expected.
```
sd-st:
```suggestion
         // only catch exceptions here as at every other occasion it's not expected.
         catch (EndOfStreamException)
```

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+        if (!Crc32ChecksumValidation(preamble[..8].Span, preamble[8..].Span))
+        {
+            throw new InvalidDataException(
+                $"The preamble at position {source.Position} is invalid"
+            );
+        }
+
+        if (
+            !BinaryPrimitives.TryReadInt32BigEndian(preamble[0..4].Span, out var totalLength)
+            || !BinaryPrimitives.TryReadInt32BigEndian(preamble[4..8].Span, out var headerLength)
```
sd-st, line 54: Just curious, are these actually signed values? Seems like it would be more sensible if they were unsigned but I'm not sure what Amazon does in practice

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
```diff
@@ -0,0 +1,169 @@
...
+            throw new InvalidDataException($"The preamble lengths are invalid");
+        }
+
+        // we don't care about headers so skip them
+        Memory<byte> header = new byte[headerLength];
+        await source.ReadExactlyAsync(header, cancellationToken).ConfigureAwait(false);
+
+        // total length is without the preamble (8bytes) + preamble crc (4bytes) + headers but do not take the message crc (4 bytes)
+        var messageLength = totalLength - 12 - headerLength - 4;
+#pragma warning disable IDE0063 // Use simple 'using' statement
```
sd-st: Just curious, what does this code look like without this? I would prefer we don't disable warnings if possible but I won't die on that hill if it's really ugly

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,108 @@
+#if !NET
+using System.Buffers;
+using System.Runtime.InteropServices;
+
+/// <summary>
+/// Contains Shims for .Net 9 available methods to compile on an NetStandard2.0 target.
```
sd-st: The comment here is for .NET 9 but we're also targeting .NET 8 - are we actually using anything that's .NET 9 specific?

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,108 @@
...
+    )
+    {
+        ValueTask<int> vt = stream.ReadAtLeastAsyncCore(
+            buffer,
+            buffer.Length,
+            throwOnEndOfStream: true,
+            cancellationToken
+        );
+
+        return vt;
```
sd-st: This is a NIT but we could make this one line, possibly even an inline method

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,108 @@
...
+    private static async ValueTask<int> ReadAtLeastAsyncCore(
+        this Stream stream,
+        Memory<byte> buffer,
+        int minimumBytes,
+        bool throwOnEndOfStream,
+        CancellationToken cancellationToken
+    )
+    {
+        int totalRead = 0;
+        while (totalRead < minimumBytes)
```
sd-st, line 42: I'm curious if this is going to act as a spinloop. Like `ReadAsync` may keep on returning 0 bytes and then we'll end up burning a lot of CPU cycles.

@stephentoub Do you know if my thinking here is correct? If so, do you have any recommended alternatives?

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,108 @@
...
+        {
+            int read = await stream
+                .ReadAsync(buffer.Slice(totalRead), cancellationToken)
+                .ConfigureAwait(false);
+            if (read == 0)
```
sd-st, line 47: Hm, is this correct? Seems like we could throw early if there just happen to be no bytes to read (the stream may not be closed yet)

sd-st (changes requested):

src/Anthropic.Bedrock/LocalShims.cs
stephentoub: If ReadAsync returns 0, it'll exit out of the loop with the throw or return below.

It's possible an invalid stream could cause problems, e.g. by erroneously returning negative values sometimes, but such invalid stream implementations could also do arbitrarily bad stuff in their implementations separate from this consuming code.

src/Anthropic.Tests/Services/Beta/Messages/BatchServiceTest.cs
JPVenson: bedrock really only supports two endpoints. I asked damian about it and he confirmed it. We also explicitly check for that in the code

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
JPVenson: IIRC i tested that first and it seemed to be an signed int

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
JPVenson: the c# compiler allows this syntax:

`using var bodyData = MemoryPool<byte>.Shared.Rent(messageLength)`

which it then will define the scope of the using itself, but that does only work if the compiler can actually know when the usage of the object ends. I am calling another method just below and then do not use the bodyData anymore but the scope __needs__ to exist until the very end of the method. I am not certain the compiler does understand this and its not intuitive to the reader that this is the case so thats why i changed it. But the compiler doesn't know that, thats why its warning here and we have to disable that warning to get the linter to stop complaining

src/Anthropic.Bedrock/LocalShims.cs
sd-st: But it shouldn't exit on 0 should it? It's possible for an open stream to return 0 bytes and then later on return more bytes from the stream

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
sd-st: It looks like you're right though AWS's documentation is super unclear. I looked at the .NET AWS SDK to verify

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
sd-st: Are they not syntactic sugar for the same thing? My understanding was that it would dispose it once the `bodyData` variable went out of scope

https://stackoverflow.com/questions/75012844/when-is-a-using-var-disposed-is-it-out-of-scope-as-soon-as-possible-or-at-the

src/Anthropic.Bedrock/LocalShims.cs
stephentoub: No, 0 means eof. A call to read will block until either there's at least one byte to return or it hits eof. A call to readasync will return a task that similarly won't complete until there's either at least one byte to return or it hits eof. Once read returns 0 all subsequent calls to it must return 0.

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
JPVenson: they are doing fundamentially the same thing yes.

```
using(var x = ...)
{
var f = x.Read();
x.Write();

console.WriteLine(f);
}
```

is the same as

```
using var x = ...;

var f = x.Read();
x.Write();
console.WriteLine(f);

```

The compiler will infer the usage of `x` and build the scope of the using itself. That however does only work if the compiler really knows how to infer the scope and where its used. In our scenario the `f` variable extends the scope of the using. For the example above its irrelevant if `f` is in scope or not, for our code it is though:

```
using (var bodyData = MemoryPool<byte>.Shared.Rent(messageLength))
{
	Memory<byte> messageSpan = bodyData.Memory[0..messageLength];
...
    var result = await Parse(new ReadOnlySequence<byte>(messageSpan), cancellationToken)
        .ConfigureAwait(false);
    return (result, true);
}
```

`messageSpan` relies on the scope of the using being extended until the return before its disposed but bodyData is only used directly up until the very first instruction. So its very unintuitive to the reader that the scope MUST exist until we return which a using statement alone does not convay. Also i am not too sure how good the compiler can infer that.

stephentoub: > The compiler will infer the usage of x and build the scope of the using itself.

The scope of a using is until the end of the parent scope. It's not inferred based on the last place the used variable is referenced. If I have:
```
using var x =...
...
Use(x);
... // lots more code that doesn't touch x
```
x is not going to be disposed of until after that "lots more code".

Changing your using block to a using declaration will have identical results.

src/Anthropic.Bedrock/LocalShims.cs
sd-st: You're right, my bad!

src/Anthropic.Bedrock/AwsEventStreamHelpers.cs
sd-st: > So its very unintuitive to the reader that the scope MUST exist until we return which a using statement alone does not convay

Imo a comment would be good to document this

stephentoub: I'm not understanding the objection to a using declaration here, though. It's understood that the scope of the using extends to the end of the containing scope. Vast amounts of code depend on that.

sd-st: I'm okay either way, it just seems a little weird to me to disable the lint if they're about the same in terms of readability

stephentoub: > it just seems a little weird to me to disable the lint if they're about the same in terms of readability

Agreed. And personally I find the using declaration easier in terms of readability. Less fuss, less indentation, etc.

JPVenson: > Changing your using block to a using declaration will have identical results.

That i just missremembered. I still prefer the normal using statement because its more verbose but in light of ^ that info i will just change it back.

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,104 @@
+#if !NET
+using System.Buffers;
+using System.Runtime.InteropServices;
+
+/// <summary>
+/// Contains Shims for .Net 8 available methods to compile on an NetStandard2.0 target.
```
sd-st:
```suggestion
/// Contains Shims for .NET 8 available methods to compile on an NetStandard2.0 target.
```

sd-st (commented):
One last NIT. Also going to test this out locally

examples/MessagesExample/Program.cs
```diff
@@ -7,6 +7,15 @@
 // For using the Foundry client, use this instead
 // AnthropicFoundryClient client = new(new AnthropicFoundryApiKeyCredentials("API-TOKEN", "RESOURCE-NAME"));
 
+// For using the Bedrock client, use this instead
+
+// using Anthropic.Bedrock;
```
sd-st: Looks like for this to work we also need to add a reference to `Anthropic.Bedrock.csproj`

Tbh I am inclinced to split this into 3 examples

src/Anthropic.Bedrock/AnthropicBedrockCredentialsHelper.cs
```diff
@@ -0,0 +1,87 @@
...
+/// // if (creds == null) { /* handle missing credentials */ }
+/// </example>
+public static class AnthropicBedrockCredentialsHelper
+{
+    public static async ValueTask<IAnthropicBedrockCredentials?> FromEnv()
+    {
+        const string ENV_API_KEY = "AWS_BEARER_TOKEN_BEDROCK";
+        const string ENV_REGION = "AWS_REGION";
+
+        var region = Environment.GetEnvironmentVariable(ENV_REGION); // this is a bit redundant as FallbackRegionFactory checks for the same value too, but its aligned with the behavior of the java sdk
```
sd-st: I think it probably makes sense to remove this comment

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,280 @@
...
+
+    private readonly IAnthropicBedrockCredentials _bedrockCredentials;
+    private readonly Lazy<IAnthropicClientWithRawResponse> _withRawResponse;
+
+    /// <summary>
+    /// Creates a new Instance of the <see cref="AnthropicBedrockClient"/>.
+    /// </summary>
+    /// <param name="bedrockCredentials">The credential Provider used to authenticate with the AWS Bedrock service.</param>
+    public AnthropicBedrockClient(IAnthropicBedrockCredentials bedrockCredentials)
+        : base()
```
sd-st, line 24: I think this should be a 0-parameter method and `BedrockCredentials` should have an init accessor to set it (with it defaulting to `FromEnv` if not set). This would be more consistent with the normal client

examples/MessagesExample/Program.cs
```diff
@@ -7,6 +7,15 @@
 // For using the Foundry client, use this instead
 // AnthropicFoundryClient client = new(new AnthropicFoundryApiKeyCredentials("API-TOKEN", "RESOURCE-NAME"));
 
+// For using the Bedrock client, use this instead
```
sd-st: This doesn't seem to work on my end:

```
AnthropicBedrockClient client = new(new AnthropicBedrockPrivateKeyCredentials()
{
    ApiAccessKey = "",
    ApiSecret = "",
    Region = "",
});


MessageCreateParams parameters = new()
{
    MaxTokens = 2048,
    Messages =
    [
        new() { Content = "Tell me a story about building the best SDK!", Role = Role.User },
    ],
    Model = "claude-sonnet-4-5",
};

var response = await client.Messages.Create(parameters);


var message = string.Join(
    "",
    response.Content
    .Select(message => message.Value)
    .OfType<TextBlock>()
        .Select((textBlock) => textBlock.Text)
);

Console.WriteLine(message);
```

I'm getting `Unhandled exception. Anthropic.Exceptions.AnthropicInvalidDataException: 'content' cannot be absent`

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
JPVenson: its not that easy. The loading from the env requires async method and constructores are not async capable. We would need to use the GetResult() antipattern to load that which i would avoid if possible

sd-st: I think I don't understand why this needs to be async at all. The normal client accesses environment variables and it does so in a sync fashion

JPVenson: the system identity is managed by the AWS lib in use and the "GetCredentialsAsync" method is used to obtain those

JPVenson: thats nessesary for feature parity with the other SDKs

sd-st: Okay, thanks

examples/MessagesExample/Program.cs
JPVenson: split

examples/MessagesExample.Bedrock/MessagesExample.Bedrock.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net9.0</TargetFramework>
```
sd-st: We should change this to 8.0 in the examples, since we want to support the oldest non-EOL version

sd-st: This is admittedly a bit of a nit since it's not super critical for the examples

examples/MessagesExample.Bedrock/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+    new AnthropicBedrockApiTokenCredentials() { BearerToken = "API-TOKEN", Region = "REGION-NAME" }
+);
+MessageCreateParams parameters = new()
+{
+    MaxTokens = 2048,
+    Messages =
+    [
+        new() { Content = "Tell me a story about building the best SDK!", Role = Role.User },
+    ],
+    Model = "claude-sonnet-4-5",
```
sd-st: I think this will need to be changed

examples/MessagesExample.Bedrock/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+using Anthropic.Models.Messages;
+
+// Create a new AnthropicBedrockClient
+// Options for the AnthropicBedrockClient constructor parameter (implements IAnthropicBedrockCredentials):
+
+// - AnthropicBedrockApiTokenCredentials — bearer-token auth (sets Authorization: bearer <token>) plus Region.
+// - AnthropicBedrockPrivateKeyCredentials — access-key/secret-key auth (for AWS-signed requests) plus Region.
+// - AnthropicBedrockCredentialsHelper.FromEnv() — helper that returns either of the above by reading AWS env vars / default credential chain.
+// - Any custom type that implements IAnthropicBedrockCredentials — implement Region and Apply(HttpRequestMessage) for custom auth.
+AnthropicBedrockClient client = new(
```
sd-st, line 16: I think defaulting to `FromEnv` and listing out the environment variables is probably best; wouldn't want someone to accidentally leave their creds in here when they open a PR

examples/MessagesExample.Foundry/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+);
+
+MessageCreateParams parameters = new()
+{
+    MaxTokens = 2048,
+    Messages =
+    [
+        new() { Content = "Tell me a story about building the best SDK!", Role = Role.User },
+    ],
+    Model = "claude-sonnet-4-5",
```
sd-st, line 21: Same thing; probably need to check that this works if you don't mind

sd-st: It's weird but the model names are different per-provider. Copying Java is probably best

examples/MessagesExample.Bedrock/Program.cs
sd-st: FYI @JPVenson I think this was missed

examples/MessagesExample.Bedrock/Program.cs
sd-st: Same thing here

examples/MessagesExample.Foundry/Program.cs
JPVenson: That model works for foundry. Its in the unittests and that was tested with this model extensively

examples/MessagesExample.Bedrock/Program.cs
JPVenson: Changed to `global.anthropic.claude-haiku-4-5-20251001-v1:0`

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,288 @@
...
+            throw new AnthropicInvalidDataException(
+                "Request is missing required path segments. Expected > 1 segments found none."
+            );
+        }
+
+        if (requestMessage.RequestUri.Segments.Length < 1)
+        {
+            throw new AnthropicInvalidDataException(
+                "Request is missing required path segments. Expected > 1 segments found none."
+            );
```
sd-st:
```suggestion
            throw new AnthropicInvalidDataException(
                "Request is missing required path segments. Expected at least 1 segment, found none."
            );
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        if (requestMessage.RequestUri.Segments[1].Trim('/') != "v1")
+        {
+            throw new AnthropicInvalidDataException(
+                $"Request is missing required path segments. Expected [0] segment to be 'v1' found {requestMessage.RequestUri.Segments[0]}."
```
sd-st:
```suggestion
                $"Request is missing required path segments. Expected [0] segment to be 'v1', found {requestMessage.RequestUri.Segments[0]}."
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,288 @@
...
+                requestMessage.Headers.TryAddWithoutValidation("Accept-Encoding", "gzip");
+            }
+        }
+
+        await _credentials.Apply(requestMessage).ConfigureAwait(false);
+    }
+
+    private static void ValidateRequest(HttpRequestMessage requestMessage)
+    {
+        if (requestMessage.RequestUri is null)
```
sd-st: Can combine this into the below if statement

src/Anthropic.Bedrock/LocalShims.cs
```diff
@@ -0,0 +1,104 @@
+#if !NET
+using System.Buffers;
+using System.Runtime.InteropServices;
+
+/// <summary>
+/// Contains Shims for .NET 8 available methods to compile on an NetStandard2.0 target.
```
sd-st:
```suggestion
/// Contains Shims for .NET 8 available methods to compile on a NetStandard2.0 target.
```

sd-st (commented):
Just some minor comments which I can commit, and I'm going to give this a test locally to make sure it works!

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,288 @@
...
+                "Request is missing required path segments. Expected > 1 segments found none."
+            );
+        }
+
+        if (requestMessage.RequestUri.Segments.Length < 1)
+        {
+            throw new AnthropicInvalidDataException(
+                "Request is missing required path segments. Expected > 1 segments found none."
+            );
+        }
```
sd-st:
```suggestion
        if (requestMessage.RequestUri is null ||
            requestMessage.RequestUri.Segments.Length < 1)
        {
            throw new AnthropicInvalidDataException(
                "Request is missing required path segments. Expected > 1 segments, found none."
            );
        }
```

src/Anthropic.Bedrock/AnthropicBedrockClient.cs
```diff
@@ -0,0 +1,282 @@
...
+            }
+        }
+
+        await _credentials.Apply(requestMessage).ConfigureAwait(false);
+    }
+
+    private static void ValidateRequest(HttpRequestMessage requestMessage)
+    {
+        if (requestMessage.RequestUri is null ||
+            requestMessage.RequestUri.Segments.Length < 1)
```
sd-st: haha yes github suggestions are the most miserable way to edit code


```suggestion
        if (requestMessage.RequestUri is null || requestMessage.RequestUri.Segments.Length < 1)
```

sd-st:
Works, merging! Thanks for this (:

PR #81, opened 2025-12-23
Fix: Dispose HttpResponseMessage after IAsyncEnumerable iteration completes

### LINKED ISSUE #80: CreateStreaming disposes HttpResponseMessage before IAsyncEnumerable is fully consumed
Problem
The MessageService.CreateStreaming() method uses using var response = ... which disposes the HttpResponseMessage before the IAsyncEnumerable<RawMessageStreamEvent> is fully iterated.

This works in production (streaming from real API), but breaks unit tests where consumers mock HttpMessageHandler with pre-recorded responses. The lazy evaluation of IAsyncEnumerable causes ObjectDisposedException because the response is already disposed when iteration begins.

Reproduction
Any unit test that:

1. Mocks HttpMessageHandler to return a streaming response
2. Calls CreateStreaming()
3. Iterates over the result

Cozmopolit:
Summary
Fixes #80

The CreateStreaming method was disposing the HttpResponseMessage before the IAsyncEnumerable was fully consumed, causing ObjectDisposedException in unit tests with mocked HTTP responses.

Changes
Replaced using var response with explicit try-finally pattern to ensure the response is disposed only after iteration completes.

Before:

using var response = await this._client...Execute(request, cancellationToken);
await foreach (var message in Sse.Enumerate<RawMessageStreamEvent>(response.Message, cancellationToken))
{
    yield return message;
}
After:

var response = await this._client...Execute(request, cancellationToken);
try
{
    await foreach (var message in Sse.Enumerate<RawMessageStreamEvent>(response.Message, cancellationToken))
    {
        yield return message;
    }
}
finally

Why this matters
Production: No change in behavior - works as before
Unit tests: Consumers can now properly test streaming with mocked HttpMessageHandler

### DIFF
--- a/src/Anthropic/Services/Beta/MessageService.cs
+++ b/src/Anthropic/Services/Beta/MessageService.cs
@@ -88,7 +88,11 @@ public sealed class MessageService : global::Anthropic.Services.Beta.IMessageSer
             Method = HttpMethod.Post,
             Params = parameters,
         };
-        using var response = await this
+        // Note: We use try-finally instead of 'using' to ensure the response is not disposed
+        // before the IAsyncEnumerable is fully consumed. With 'using', the response would be
+        // disposed when the method returns the IAsyncEnumerable, not when iteration completes.
+        // See: https://github.com/anthropics/anthropic-sdk-csharp/issues/80
+        var response = await this
             ._client.WithOptions(options =>
                 options with
                 {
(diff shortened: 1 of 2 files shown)


TomerAberbach:
I'm not sure I understand how this is different. Isn't `using var` just syntactic sugar for exactly what you've written? `using var` will only result in disposal at the end of the function, and the end of the function is only reached after the SSE messages are fully iterated, so I would expect the behavior to be identical.

What am I missing here?

Cozmopolit:
@TomerAberbach 

Freck. So, after months of AI-assisted coding, AI slop happened publically to me for the first time.

The "issue" was in the test (for the Anthropic Connector for Semantic Kernel that I am writing up) - _not_ here in the SDK.

gemini-3-pro, claude-opus-4.5, gpt-5.1 and a couple more all agreed on problem analysis and the solution - which was all bogus as you correctly commented.

I am closing this PR and the related issue - I apologize deeply for wasting your time.

PR #86, opened 2026-01-04
fix: remove using statement from FileService.Download to prevent Obje…

### LINKED ISSUE #85: Files API FileService's Download method disposes the HttpResponse before returning it
When calling the `FileService`'s `Download` method and trying to access the content stream, a `ObjectDisposedException` is thrown. This is due to the `Download` method disposing the `HttpResponse` before returning it by prepending the request with `using`.  This disposes the underlying `HttpResponseMessage`, which disposes `Message.Content` (`HttpConnectionResponseContent`) and is why the call to `response.ReadAsStream(...) ` blows up with `ObjectDisposedException`.

Cozmopolit:
Fixes #85

## Summary

`FileService.Download` uses `using var response` but returns the response, disposing it before the caller can use it.

**Fix:** Remove `using`, add XML docs clarifying caller disposal responsibility.

## Contribution Checklist

- [x] Builds clean
- [x] Follows existing SDK patterns
- [x] No breaking changes
- [x] Tests pass

### DIFF
--- a/src/Anthropic/Services/Beta/FileService.cs
+++ b/src/Anthropic/Services/Beta/FileService.cs
@@ -112,7 +112,9 @@ public sealed class FileService : IFileService
             Method = HttpMethod.Get,
             Params = parameters,
         };
-        using var response = await this
+        // Note: We intentionally do not dispose the response here.
+        // The caller is responsible for disposing the returned HttpResponse.
+        var response = await this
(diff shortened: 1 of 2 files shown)


sd-st:
Hi, thanks for the PR! I've logged a ticket internally so that we can fix this as a part of codegen. I'll update #85 once we have that implemented.

PR #91, opened 2026-01-12
Allow MEAI User Agent tracking for IChatClient

rogerbarreto:
Similar to other specializations of `IChatClient` adding the additional `MEAI` in `User-Agent` improves the understanding of the `Microsoft.Extensions.AI` ecosystem and also helps to solidify it's usage and importance from the MEAI abstraction perspective.

- [MEAI.OpenAI package](https://github.com/dotnet/extensions/blob/99383237b2b86d4d85b4f5908b938e9e43c97db7/src/Libraries/Microsoft.Extensions.AI.OpenAI/RequestOptionsExtensions.cs#L32)
- [MEAI Azure Foundry persistent package](https://github.com/Azure/azure-sdk-for-net/blob/9d32be0c40d6ded048b2c85ea63f73497fb05350/sdk/ai/Azure.AI.Agents.Persistent/src/Custom/PersistentAgentsChatClient.cs#L802)


src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -111,16 +147,16 @@ void IDisposable.Dispose() { }
 
             if (serviceType == typeof(ChatClientMetadata))
             {
-                return _metadata ??= new(
+                return this._metadata ??= new(
```
stephentoub: Why did all these `this.` get added?

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -87,12 +88,47 @@ private sealed class AnthropicChatClient(
     ) : IChatClient
     {
         private const int DefaultMaxTokens = 1024;
+        private const string MeaiUserAgentHeaderKey = "User-Agent";
+
+        private static readonly IReadOnlyDictionary<string, JsonElement> s_meaiHeaderData =
```
stephentoub:
```suggestion
        private static readonly Dictionary<string, JsonElement> s_meaiHeaderData =

```

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -905,7 +941,28 @@ toolMode is AutoChatToolMode
...
+            return AddMeaiHeaders(createParams);
+        }
+
+        private static MessageCreateParams AddMeaiHeaders(MessageCreateParams createParams)
+        {
+            Dictionary<string, JsonElement> mergedHeaders = [];
+            foreach (var header in s_meaiHeaderData)
+            {
+                mergedHeaders[header.Key] = header.Value;
+            }
```
stephentoub:
```suggestion
            Dictionary<string, JsonElement> mergedHeaders = new(s_meaiHeaderData);

```

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -899,8 +899,10 @@ public async Task GetResponseAsync_WithToolCalls()
                     "description": "Get the current weather for a location",
                     "input_schema": {
                         "type": "object",
-                        "location": { "type": "string", "description": "The city and state" },
-                        "unit": { "type": "string", "description": "Temperature unit" },
+                        "properties": {
+                            "location":{"type":"string","description":"The city and state"},
+                            "unit":{"type":"string","description":"Temperature unit"}
+                        },
```
stephentoub, line 905: Were these not changed as part of the previous PR that fixed properties? How did that pass CI tests?

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -905,7 +941,28 @@ toolMode is AutoChatToolMode
...
+
+            foreach (var header in createParams.RawHeaderData)
+            {
+                mergedHeaders[header.Key] = header.Value;
+            }
+
+            return MessageCreateParams.FromRawUnchecked(
```
stephentoub, line 952: Is this appropriately resulting in a single User-Agent header that contains both the anthropic and meai values, rather than two separate User-Agent headers?

rogerbarreto: @stephentoub Yes, in the implementation, the `Anthropic` user agent is added separately, first the static `Anthropic UserAgent` then `RawHeaderData`.

image


https://github.com/anthropics/anthropic-sdk-csharp/blob/9ad3dc0393ee0e7d8ea592442f4dd12baca83dc4/src/Anthropic/Models/Messages/MessageCreateParams.cs#L449-L455

The methods use the `HttpRequestHeaders.Add` method that appends properly the headers.

src/Anthropic/AnthropicClientExtensions.cs
rogerbarreto: My editor has the local .editorconfig enforcing it when a repo doesn't have the rule set. reverted and added the rule in the repo to ignore.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
rogerbarreto: Running the tests from my latest branch from the `next` locally were failing, added the fix in a dedicated commit to this PR.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -87,12 +88,46 @@ private sealed class AnthropicChatClient(
     ) : IChatClient
     {
         private const int DefaultMaxTokens = 1024;
+        private const string MeaiUserAgentHeaderKey = "User-Agent";
+
+        private static readonly Dictionary<string, JsonElement> s_meaiHeaderData = new()
```
sd-st: Can we make this a readonly dictionary, possibly even frozen?

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1149,7 +1185,28 @@ existingSystem.Value is IReadOnlyList<BetaTextBlockParam> existingMessages
...
+            Dictionary<string, JsonElement> mergedHeaders = [];
+            foreach (var header in s_meaiHeaderData)
+            {
+                mergedHeaders[header.Key] = header.Value;
+            }
+
+            foreach (var header in createParams.RawHeaderData)
+            {
+                mergedHeaders[header.Key] = header.Value;
+            }
```
sd-st: Just curious, why do we prioritize `mergedHeaders` over `createParams.RawHeadersData` here, but in the other file it's the other way around?

sd-st: I think we could also do the `new(...)` thing in here when instantiating the dictionary (like we do in the other file) to save us one of the loops

sd-st: Also, I'm wondering if we should be combining headers rather than overwriting them?

src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
```diff
@@ -2218,4 +2218,259 @@ public async Task GetResponseAsync_WithHostedToolsAndExistingBetas_PreservesAndD
...
+                }
+            }
+            """
+        )
+        {
+            OnRequestHeaders = headers =>
+            {
+                if (headers.TryGetValues("User-Agent", out var values))
+                {
+                    capturedUserAgent = string.Join(" ", values);
```
sd-st: This `string.Join` shouldn't be needed right? I thought we were already coming user agents like this?

I think we should assert that the user agent is an array of length 1

src/Anthropic/AnthropicClientExtensions.cs
sd-st: Isn't this only done in a unit test? I think even outside of the unit test we'd want a single `User-Agent` header

sd-st (approved):
Thanks!

PR #92, opened 2026-01-13
feat(client): support `WithRawResponse` in Foundry client

JPVenson:


src/Anthropic.Foundry/AnthropicFoundryClient.cs
```diff
@@ -8,6 +8,7 @@ namespace Anthropic.Foundry;
 public class AnthropicFoundryClient : AnthropicClient
 {
     private readonly IAnthropicFoundryCredentials _azureCredentials;
+    private readonly Lazy<IAnthropicClientWithRawResponse> _withRawResponseFactory;
```
TomerAberbach:
```suggestion
    private readonly Lazy<IAnthropicClientWithRawResponse> _withRawResponse;
```

(I think this is what we call it in the other classes?)

PR #105, opened 2026-02-02
feat(client): streaming aggregators

sd-st:


src/Anthropic.Tests/Services/Beta/MessageServiceTest.cs
```diff
@@ -15,7 +50,14 @@ public async Task Create_Works(IAnthropicClient client)
             new()
             {
                 MaxTokens = 1024,
-                Messages = [new() { Content = "Hello, world", Role = Role.User }],
+                Messages =
+                [
+                    new()
+                    {
+                        Content = "Hello, world",
```
sd-st: todo: gotta fix these diffs

felixfbecker (approved):
Love the LINQ integration!

One callout: In other SDKs we have a warning (or error?) that gets triggered when you use sync requests with potentially large input/output to tell the user to use streaming instead. We probably want to add the same if it doesn't exist yet, especially now that we have the aggregator so that you can use the streaming API the same way as the sync API in a one-liner.

sd-st:
@felixfbecker Great call-out! I'll add that today

examples/Anthropic.Examples.sln
TomerAberbach, line 1: Do we need to add `StreamingAggregationBetaExample` too

src/Anthropic/Services/Beta/MessageService.cs
```diff
@@ -124,7 +124,13 @@ public async Task<HttpResponse<BetaMessage>> Create(
             ._client.WithOptions(options =>
                 options with
                 {
-                    Timeout = options.Timeout ?? TimeSpan.FromMinutes(10),
+                    Timeout =
+                        options.Timeout
+                        ?? ClientOptions.TimeoutFromMaxTokens(
+                            parameters.MaxTokens,
+                            false,
```
TomerAberbach: nit: add `paramName:` so that it's clear what this is

Similarly for other usages

src/Anthropic/SseAggregatorExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+/// Extension methods for providing easy access to Aggregators
+/// </summary>
+public static class SseAggregatorExtensions
+{
+    /// <summary>
+    /// Aggregates all messages received by the streaming event and aggregates them into a single object once the sender indicates a fully delivered stream.
+    /// </summary>
+    /// <param name="source">A enumerable as provided by the <see cref="IMessageService.CreateStreaming(Models.Messages.MessageCreateParams)"/></param>
+    /// <returns>A task that completes once all messages have been received or in the event of improper streaming and exception.</returns>
+    public static async Task<Message?> Aggregate(
```
TomerAberbach: Hmmm, why is this nullable? And similarly for the other one?

In Java we make it non-nullable because I think the only way it'd not be there is due to some malformed API stream (in which case the accumulator throws)

src/Anthropic/SseAggregatorExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+    /// <param name="aggregator">The aggregator instance that will collect messages and build the result object.</param>
+    /// <returns>An <see cref="IAsyncEnumerable{TMessage}"/> filtered by the aggregator.</returns>
+    public static IAsyncEnumerable<TMessage> CollectAsync<TMessage, TResult>(
+        this IAsyncEnumerable<TMessage> source,
+        SseAggregator<TMessage, TResult> aggregator
+    )
+    {
+        return aggregator.CollectAsync(source);
+    }
+}
```
TomerAberbach, line 68: I am confused by this method. Isn't this just directly delegating to `SseAggregator.CollectAsync`? Which is already public?

README.md
```diff
@@ -123,6 +123,13 @@ To send a request to the Anthropic API, build an instance of some `Params` class
 
 For example, `client.Messages.Create` should be called with an instance of `MessageCreateParams`, and it will return an instance of `Task<Message>`.
 
+> [!IMPORTANT]
+> We highly encourage you to use [streaming](#streaming) for longer running requests.
+
+We do not recommend setting a large `maxTokens` value without using streaming. Some networks may drop idle connections after a certain period of time, which can cause the request to fail or [timeout](#timeouts) without receiving a response from Anthropic. We periodically ping the API to keep the connection alive and reduce the impact of these networks.
```
TomerAberbach:
```suggestion
We do not recommend setting a large `MaxTokens` value without using streaming. Some networks may drop idle connections after a certain period of time, which can cause the request to fail or [timeout](#timeouts) without receiving a response from Anthropic. We periodically ping the API to keep the connection alive and reduce the impact of these networks.
```

`maxTokens` is from Java builder casing I think?

README.md
```diff
@@ -155,6 +162,44 @@ await foreach (var message in client.Messages.CreateStreaming(parameters))
...
+        else if (delta.Delta.TryPickText(out var textDelta))
+        {
+            Console.Write(textDelta.Text);
+        }
+    }
+}
+
+// And then get the full aggregated message.
+var fullMessage = await aggregator.Message();
+
```
TomerAberbach:
```suggestion
```

examples/Anthropic.Examples.sln
sd-st: hmm what do we typically do? I figured it would be basically identical but happy to add it if that's the status quo and/or you think it would be helpful to end-users

TomerAberbach: To clarify, we _do_ have a StreamingAggregationBetaExample in this PR, but it's not in this sln file (unlike the non-beta one)

sd-st: oh yes then we should definitely add it, good catch

src/Anthropic/SseAggregatorExtensions.cs
sd-st: This allows you to chain it within a LINQ tree with `CreateStreaming().CollectAsync(aggregator).Select(...)` etc

TomerAberbach (approved):

PR #108, opened 2026-02-03
(feat) Add vertex provider

JPVenson:
Adds googles `vertex` platform as supported provider
Adds Examples as own project (as discussed in the #77 pr)
Adds inclusion into tests. 
	Note: As vertex the same as bedrock does not follow the same api structure as anthropic does, testing with the internal test server isn't possible as of now. The code has been manually tested for compliance for all supported endpoints


src/Anthropic.Vertex/AnthropicVertexClientWithRawResponse.cs
```diff
@@ -0,0 +1,135 @@
...
+            await requestMessage.Content!.ReadAsStringAsync(
+#if NET
+                cancellationToken
+#endif
+            ).ConfigureAwait(false)
+        )!;
+
+        bodyContent["anthropic_version"] = JsonValue.Create(ANTHROPIC_VERSION);
+
+        var modelValue = bodyContent["model"]!;
```
sd-st: I'm generally not a huge fan of the `!` operator especially as it seems like this could technically be missing. Would be good to check that it's null and raise if it's not imo

src/Anthropic.Vertex/AnthropicVertexClientWithRawResponse.cs
```diff
@@ -0,0 +1,135 @@
...
+            await writer.FlushAsync(cancellationToken).ConfigureAwait(false);
+        }
+        contentStream.Seek(0, SeekOrigin.Begin);
+        requestMessage.Headers.TryAddWithoutValidation(
+            "content-length",
+            contentStream.Length.ToString()
+        );
+
+        var requestBuilder = new StringBuilder(
+            $"{requestMessage.RequestUri!.Scheme}://{requestMessage.RequestUri.Host}/v1/projects/{_vertexCredentials.Project}/locations/{_vertexCredentials.Region}/publishers/anthropic/models/"
```
sd-st, line 76: Same thing here

src/Anthropic.Vertex/AnthropicVertexClientWithRawResponse.cs
```diff
@@ -0,0 +1,135 @@
...
+            throw new AnthropicInvalidDataException(
+                "Request is missing required path segments. Expected > 1 segments found none."
+            );
+        }
+
+        if (requestMessage.RequestUri.Segments[1].Trim('/') != "v1")
+        {
+            throw new AnthropicInvalidDataException(
+                $"Request is missing required path segments. Expected [0] segment to be 'v1' found {requestMessage.RequestUri.Segments[0]}."
+            );
```
sd-st, line 114: can probably steal the same exception messages from bedrock since we changed them

src/Anthropic.Vertex/AnthropicVertexClientWithRawResponse.cs
```diff
@@ -0,0 +1,135 @@
...
+            && requestMessage.RequestUri.Segments[2].Trim('/') is "messages"
+            && requestMessage.RequestUri.Segments[3].Trim('/') is "count_tokens"
+        )
+        {
+            isCountEndpoint = true;
+        }
+        else
+        {
+            isCountEndpoint = false;
+        }
```
sd-st:
```suggestion
        isCountEndpoint = requestMessage.RequestUri.Segments.Length >= 4
            && requestMessage.RequestUri.Segments[2].Trim('/') is "messages"
            && requestMessage.RequestUri.Segments[3].Trim('/') is "count_tokens";
```

sd-st (commented):
Nice! Just had a few minor comments

src/Anthropic.Vertex/AnthropicVertexClientWithRawResponse.cs
JPVenson: IIRC i did add a full check for nullability in bedrock too and there the objection was that if somebody were to input wrong data it should just fail so that would be very different here then

sd-st: I mean it's going to fail either way right? Wouldn't it be better to provide a more descriptive error message?

JPVenson: well i totally agree, but its the opposite of what was decided in the other PR, oh wait let me look it up maybe that was tomer

sd-st: yeah I'm wondering if there was a miscommunication there; if you have the comment handy I'd love to take a look

JPVenson: as discussed in chat, i will readd the check as added in bedrock previously here and readd the check in bedrock alongside it

sd-st:
LGTM! Needs a rebase (which may require some small code changes; they updated their spec)

sd-st:
Also how did you test this locally with an access token/project ID/region?

JPVenson:
ahrrrrg i JUST rebased it like 30 minutes ago!

i am currently writing an doc on steps to get the nessesary google stuff setup but i can send you the test stuff on slack. The vertex example holds everything nessesary for it in its comment

sd-st:
Yeah sorry they just did a release so there were some changes :sweat_smile:

sd-st:
Tested locally. A few things to address:
- model name needs to be updated in examples
- we should add the ability to use `ANTHROPIC_VERTEX_PROJECT_ID` instead of specifying the project id in the code
- we should document in the example the fact that you can configure auth using the respective environment variables (`CLOUD_ML_REGION`, `ANTHROPIC_VERTEX_PROJECT_ID`, `VERTEX_ACCESS_TOKEN`)

JPVenson:
@sd-st 

- Model name updated to `claude-3-7-sonnet@20250219`
- Added `DefaultAnthropicVertexCredentials.FromEnvAsync` in line with foundry and bedrock
- Added additional documentation on the new method of variables and references to google documentation

examples/MessagesExample.Vertex/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+
+// or you can load the credentials from your system after you set it with the necessary environment variables by calling
+// var client = new AnthropicVertexClient(DefaultAnthropicVertexCredentials.FromEnvAsync());
+
+// <code>
+// ANTHROPIC_VERTEX_PROJECT_ID=your_project_id
+// CLOUD_ML_REGION=region_name
+// VERTEX_ACCESS_TOKEN=vertex_access_token
+// </code>
+// are the main variables you can set but there are more options. Consult the methods documentation for more info.
```
sd-st:
```suggestion
// The main variables you can set are below. There are more options available; consult the method's documentation for more info.
// <code>
// ANTHROPIC_VERTEX_PROJECT_ID=your_project_id
// CLOUD_ML_REGION=region_name
// VERTEX_ACCESS_TOKEN=vertex_access_token
// </code>
```

examples/MessagesExample.Vertex/Program.cs
```diff
@@ -0,0 +1,51 @@
...
+var client = new AnthropicVertexClient(new AnthropicVertexCredentials(null, "YourProjectId", GoogleCredential.FromJson(
+"""
+{
+    ServiceAccount JSON
+}
+""").CreateScoped("https://www.googleapis.com/auth/cloud-platform")));
+*/
+
+// or you can load the credentials from your system after you set it with the necessary environment variables by calling
+// var client = new AnthropicVertexClient(DefaultAnthropicVertexCredentials.FromEnvAsync());
```
sd-st:
```suggestion
// var client = new AnthropicVertexClient(await DefaultAnthropicVertexCredentials.FromEnvAsync());
```

sd-st:
Thanks!

PR #118, opened 2026-02-10
Update to M.E.AI.Abstractions 10.4.0 and update with latest MEAI/Anthropic features

stephentoub:
This also lets us map ChatOptions.Reasoning.

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -2709,6 +2709,456 @@ public abstract class AnthropicClientExtensionsTestsBase
         Assert.Null(response.FinishReason);
     }
 
+    [Theory]
+    [InlineData(ReasoningEffort.Low, 1024)]
+    [InlineData(ReasoningEffort.Medium, 8192)]
+    [InlineData(ReasoningEffort.High, 16384)]
+    [InlineData(ReasoningEffort.ExtraHigh, 32768)]
+    public async Task GetResponseAsync_WithReasoningEffort_SetsThinkingEnabled(
+        ReasoningEffort effort,
+        int expectedBudgetTokens
+    )
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: $$"""
+            {
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Think carefully"
+                    }]
+                }],
+                "max_tokens": 100000,
+                "thinking": {
+                    "type": "enabled",
+                    "budget_tokens": {{expectedBudgetTokens}}
+                }
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_reasoning_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Here is my response"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 10,
+                    "output_tokens": 20
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        ChatOptions options = new()
+        {
+            MaxOutputTokens = 100000,
+            Reasoning = new() { Effort = effort },
+        };
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Think carefully",
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithReasoningEffortNone_SetsThinkingDisabled()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Simple question"
+                    }]
+                }],
+                "max_tokens": 1024,
+                "thinking": {
+                    "type": "disabled"
+                }
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_reasoning_02",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Quick answer"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 10,
+                    "output_tokens": 5
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        ChatOptions options = new() { Reasoning = new() { Effort = ReasoningEffort.None } };
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Simple question",
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithReasoningEffort_ClampsBudgetToExplicitMaxTokens()
+    {
+        // High effort maps to 16000, but caller explicitly set max_tokens to 5000,
+        // so budget should clamp to 4999.
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Think carefully"
+                    }]
+                }],
+                "max_tokens": 5000,
+                "thinking": {
+                    "type": "enabled",
+                    "budget_tokens": 4999
+                }
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_reasoning_03",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Response"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 10,
+                    "output_tokens": 15
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        ChatOptions options = new()
+        {
+            MaxOutputTokens = 5000,
+            Reasoning = new() { Effort = ReasoningEffort.High },
+        };
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Think carefully",
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithReasoningEffort_SkipsThinkingWhenExplicitMaxTokensTooSmall()
+    {
+        // Medium effort maps to 10024, but caller explicitly set max_tokens to 1024,
+        // so after clamping budget would be 1023 which is < 1024 minimum. Thinking is skipped.
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Think carefully"
+                    }]
+                }],
+                "max_tokens": 1024
+            }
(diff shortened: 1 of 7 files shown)


sd-st:
@stephentoub Hey, sorry about the delay on this. Would you be able to rebase?

stephentoub:
> @stephentoub Hey, sorry about the delay on this. Would you be able to rebase?

@sd-st done, thanks

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -923,6 +923,57 @@ toolMode is AutoChatToolMode
...
+                        thinkingConfig = new(new ThinkingConfigDisabled());
+                    }
+                    else
+                    {
+                        long? budgetTokens = reasoning.Effort switch
+                        {
+                            ReasoningEffort.Low => 1024,
+                            ReasoningEffort.Medium => 8192,
+                            ReasoningEffort.High => 16000,
+                            ReasoningEffort.ExtraHigh => 32000,
```
sd-st: Curious why these aren't also base 2?

sd-st: Also FYI there's a limit to tokens on the non-streaming endpoint: https://github.com/anthropics/anthropic-sdk-csharp/blob/main/src/Anthropic/Core/ClientOptions.cs#L114

It's purely client side so we could override it in the extensions. Or we could use the streaming endpoint if that's possible (preferred)

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -923,6 +923,57 @@ toolMode is AutoChatToolMode
...
+                            _ => null,
+                        };
+
+                        if (budgetTokens is { } budget)
```
sd-st, line 1356: Any reason not to do something like this?

```suggestion
                        if (budgetTokens != null)
```

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -923,6 +923,57 @@ toolMode is AutoChatToolMode
                         createParams = createParams with { ToolChoice = toolChoice };
                     }
                 }
+
+                if (createParams.Thinking is null && options.Reasoning is { } reasoning)
```
sd-st, line 1338: As an aside this method is getting pretty unwieldy. May be worth refactoring at some point

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -923,6 +923,57 @@ toolMode is AutoChatToolMode
...
+                        if (budgetTokens is { } budget)
+                        {
+                            // Anthropic requires budget_tokens >= 1024 and budget_tokens < max_tokens.
+                            if (createParams.MaxTokens <= budget)
+                            {
+                                if (options.MaxOutputTokens is not null)
+                                {
+                                    // Caller explicitly set MaxOutputTokens. Clamp the budget to fit,
+                                    // and skip thinking if it can't meet the minimum.
+                                    budget = createParams.MaxTokens - 1;
```
sd-st, line 1366: Is this going to cause thinking to take up the entire budget? Is it possible there'll be nothing left over for the response?

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -923,6 +923,57 @@ toolMode is AutoChatToolMode
...
+                                    // to accommodate the thinking budget plus room for output.
+                                    createParams = createParams with
+                                    {
+                                        MaxTokens = budget + _defaultMaxTokens,
+                                    };
+                                }
+                            }
+
+                            thinkingConfig =
+                                budget >= 1024 ? new(new ThinkingConfigEnabled(budget)) : null;
```
sd-st: If we don't end up using the thinking config I think probably we don't want to change `createParams.MaxTokens` either

sd-st (changes requested):

src/Anthropic/AnthropicClientExtensions.cs
stephentoub: > Curious why these aren't also base 2?

No good reason. I'll change it.

> Also FYI there's a limit to tokens on the non-streaming endpoint. It's purely client side so we could override it in the extensions. Or we could use the streaming endpoint if that's possible (preferred)

Why does the client library add such a limitation? That's suspect.

src/Anthropic/AnthropicClientExtensions.cs
stephentoub: `budget` here will be `long` whereas `budgetTokens` is `long?`. I wanted to do the null check and then get the `long` value for subsequent use without having to use `.Value` on each access.

src/Anthropic/AnthropicClientExtensions.cs
stephentoub: budget is a maximum, not a guaranteed amount. The model balances thinking vs response within the max tokens envelope. And when MaxOutputTokens is too small (budget clamped below 1024), thinking is skipped entirely.

src/Anthropic/AnthropicClientExtensions.cs
sd-st: Given that the non-streaming endpoints may wait for a few minutes with no bytes being sent, there are issues that can happen here especially if on a flaky network. Users should generally have a much better time on the streamining endpoints for long responses. As a result we warn the user if they attempt to use the non-streaming endpoints with a large number of tokens. This check does get disabled if the timeout is specified explicitly, so the user can override it if they really want to.

stephentoub: Ok, thanks. My $.02, that's a very heavy-handed, opinionated thing for the client to be doing. If the service is happy to allow a larger value, I suggest the client should be as well. Do the other Anthropic clients in other languages apply a similar client-side policy?

sd-st: Yes, they do

stephentoub: Ok, then I understand doing it for consistency. Still seems like a strange choice, but I get following suit. Thanks.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -2709,6 +2709,456 @@ public async Task GetResponseAsync_WithNullFinishReason()
...
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithReasoningEffort_SkipsThinkingWhenExplicitMaxTokensTooSmall()
+    {
+        // Medium effort maps to 10024, but caller explicitly set max_tokens to 1024,
```
sd-st:
```suggestion
        // Medium effort maps to 8192, but caller explicitly set max_tokens to 1024,
```

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -2709,6 +2709,456 @@ public async Task GetResponseAsync_WithNullFinishReason()
...
+            options,
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
+    }
+
+    [Fact]
+    public async Task GetResponseAsync_WithReasoningEffort_ClampsBudgetToExplicitMaxTokens()
+    {
+        // High effort maps to 16000, but caller explicitly set max_tokens to 5000,
```
sd-st:
```suggestion
        // High effort maps to 16384, but caller explicitly set max_tokens to 5000,
```

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -190,10 +190,25 @@ out List<TextBlockParam>? systemMessages
...
+                        Timeout = ClientOptions.TimeoutFromMaxTokens(
+                            createParams.MaxTokens,
+                            isStreaming: true,
+                            createParams.Model
+                        ),
+                    }
+                );
+            }
+
+            var createResult = await messageService.Create(createParams, cancellationToken);
```
sd-st, line 547: Could we use the streaming endpoint with the message aggregator helper instead of bypassing the timeout check actually?

sd-st (commented):
FYI the build is failing but that's unrelated to your changes and a rebase should fix it

src/Anthropic/AnthropicClientExtensions.cs
stephentoub: Are you asking if GetResponseAsync could instead be rewritten in terms of GetStreamingResponseAsync?

sd-st: Yes exactly

stephentoub: It _can_. The question is whether it _should_. The main downsides to that are:
1. The underlying RawRepresentation will be different, since e.g. ChatResponse.RawRepresentation won't have a single Anthropic object to return.
2. It's more expensive, getting back lots of SSE events, each of which is materialized into the various objects models, only to then be coalesced into the single non-streaming version.
3. If there are any discrepancies in what the service returns for the two variants, it'll manifest.

Of course it has the benefit of less client code to maintain.

I suggest we not make such a change in this PR. If you'd like, I can prepare a separate one to show you what it would look like, and you can decide whether to take it or not?

sd-st: Alright, sure.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
sd-st: This still seems to exist fyi

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
sd-st: same thing here

stephentoub:
Fixed the two comments (you commented on three but I'm not seeing what you were referring to in one of them)

sd-st:
> Fixed the two comments (you commented on three but I'm not seeing what you were referring to in one of them)

Hm I think I only commented on 2? Where's the third one?

stephentoub:
> > Fixed the two comments (you commented on three but I'm not seeing what you were referring to in one of them)
> 
> Hm I think I only commented on 2? Where's the third one?

I have no idea. Maybe GH was playing tricks on me :)

sd-st:
The diff is enormous now; does this need a rebase?

stephentoub:
> does this need a rebase?

I don't think so?

> The diff is enormous now

 A lot of that is tests moving down from the beta-specific test file to the base test file that applies to both IChatClient implementations. Features that were previously beta-specific were promoted to stable by Anthropic, so those features were added into the stable IChatClient as well, and the tests promoted to apply to both.

(As an aside, a lot of the implementation between the two IChatClient implementations could be shared if it weren't for the exact same "types" in the two surface areas having different C# type representations for them. It'd be nice, both for the IChatClient implementations and for consumers of the library in general, if the code generation could be done in such a way that the beta surface area only needed different types when there was actually something different to represent.)

sd-st:
> (As an aside, a lot of the implementation between the two IChatClient implementations could be shared if it weren't for the exact same "types" in the two surface areas having different C# type representations for them. It'd be nice, both for the IChatClient implementations and for consumers of the library in general, if the code generation could be done in such a way that the beta surface area only needed different types when there was actually something different to represent.)

The main issue with doing this in the general case is that it means adding a field to one type and not the other would result in a breaking change (the types would have to be split back apart)

stephentoub:
> The main issue with doing this in the general case is that it means adding a field to one type and not the other would result in a breaking change (the types would have to be split back apart)

Probably worth a separate discussion. Seems like there should be ways to address that, e.g. using extension blocks in the beta namespace to add the necessary public methods/properties.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -999,44 +1040,16 @@ createParams.OutputFormat is null
                                 break;
 
                             case AIFunctionDeclaration af:
-                                Dictionary<string, JsonElement> properties = [];
```
sd-st, line 1002: Why was this logic removed?

stephentoub: It was effectively moved into the schema transform and improved there, matching the same transformations applied in both the typescript and python anthropic client libraries.

sd-st (approved):
Thanks for your contribution! In the future I think it would be good to split up big PRs so that they're easier to review. Also FYI @cameron-mcateer there's going to be a bit of duplication between this PR and your structured outputs work

stephentoub:
> Thanks for your contribution! In the future I think it would be good to split up big PRs so that they're easier to review

Sounds good. Will try :)

PR #134, opened 2026-02-25
added first WIP draft of tool runners

JPVenson:


src/Anthropic/Anthropic.csproj
```diff
@@ -12,15 +12,18 @@
...
+    <PackageReference Include="System.Net.ServerSentEvents" Version="10.0.1" />
+    <PackageReference Include="Microsoft.Extensions.AI.Abstractions" Version="10.2.0" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="8.0.0" PrivateAssets="All" />
   </ItemGroup>
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Collections.Immutable" Version="8.0.0"/>
+    <PackageReference Include="System.Collections.Immutable" Version="8.0.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Tool/" />
```
sd-st, line 27: Just wondering why this is in a separate directory? Can probably put it in the helpers directory that we have already, similar to Java

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+    /// <summary>
+    /// Obtains a <see cref="Message"/> that is generated using client side tooling.
+    /// </summary>
+    /// <param name="messageService">The <see cref="IMessageService"/>.</param>
+    /// <param name="messageCreateParams">The <see cref="MessageCreateParams"/>.</param>
+    /// <param name="configureTools">The callback for registering tools.</param>
+    /// <param name="cancellationToken">The <see cref="CancellationToken"/></param>
+    /// <returns>A task that completes when all tools have been executed and a final message has been obtained.</returns>
+    /// <exception cref="InvalidOperationException"></exception>
+    public static async Task<Message> CreateWithToolLoop(
```
sd-st, line 29: I think we probably don't need an extension method for this. Instead, we can change the `MessageCreateParams` class itself to add the ability to specify tools. This is what Java does (although it's a little bit different because Java uses the builder pattern, while we do not)

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+        CancellationToken cancellationToken = default
+    )
+    {
+        var toolRegistry = configureTools(new());
+
+        messageCreateParams = messageCreateParams with
+        {
+            Tools =
+            [
+                .. messageCreateParams.Tools ?? [], //TODO: does this make sense because user may never be able to invoke anything here. Dynamic bindings of user code?
```
sd-st, line 42: I think we probably don't need to support the case where the user also specifies `Tool` objects

src/Anthropic/Tool/ToolRunnerExtensions.cs
sd-st: FYI this part of the Java readme is a pretty good summary of what we want to build: https://github.com/anthropics/anthropic-sdk-java/tree/main?tab=readme-ov-file#tool-use-with-json-schemas

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+                    }
+                )),
+            ],
+        };
+
+        var result = await messageService
+            .Create(messageCreateParams, cancellationToken)
+            .ConfigureAwait(false);
+
+        //TODO clarify with tomer/stephen why all examples do this one by one instead of running all tools at once
```
sd-st, line 58: Just to be clear, the LLM may or may not choose to use some (or all) of the tools, and it may use the output of one tool to inform the input to another tool. It may even call the same tool multiple times (with different arguments). So we can't really preempt the calls

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+        while (result.StopReason == StopReason.ToolUse)
+        {
+            var toolUseRequest =
+                result.Content.Select(e => e.Value).OfType<ToolUseBlock>().LastOrDefault()
+                ?? throw new InvalidOperationException(
```
sd-st, line 63: Curious if there's a good `Anthropic*Exception` we can use here? We generally like to throw these when possible so that the user can catch them all at once if they desire, and if this happens it would be a bug on the server side

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+                    "Stop reason indicates Tool Usage request but no ToolUse block was found."
+                );
+            toolRegistry.LocalTools.TryGetValue(toolUseRequest.Name, out var toolRunner);
+
+            JsonDocument toolUseInputArguments = null!; //TODO build input doc
+            var toolUseResult = toolRunner is null
+                ? null
+                : await toolRunner.Invoke(toolUseInputArguments).ConfigureAwait(false);
```
sd-st, line 71: NIT

```suggestion
            var toolUseResult = toolRunner && await toolRunner.Invoke(toolUseInputArguments).ConfigureAwait(false);
```

I think this works? But correct me if I'm wrong. I forget what C#'s short circuiting story is like

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+
+            messageCreateParams = messageCreateParams with
+            {
+                Messages =
+                [
+                    .. messageCreateParams.Messages.Except(
+                        messageCreateParams.Messages.Where(e =>
+                            e.Content.Value is IReadOnlyList<ContentBlockParam> blockParams
+                            && blockParams.Any(f => f.ToolUseID == toolUseRequest.ID)
+                        )
```
sd-st, line 81: FYI we do want to carry the existing messages along (including the tool call request) since that's what makes up the context/state of the LLM

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+                    new MessageParam()
+                    {
+                        Content = new MessageParamContent(
+                            [
+                                new ToolResultBlockParam()
+                                {
+                                    ToolUseID = toolUseRequest.ID,
+                                    Content = new(
+                                        toolUseResult?.ToString()
+                                            ?? $"Error: no result from tool '{toolUseRequest.Name}'"
```
sd-st, line 92: I think we'll probably want to do json serialization here

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+public class ToolRegistration
+{
+    public Dictionary<string, IToolRegistration> LocalTools { get; } =
+        new(StringComparer.InvariantCultureIgnoreCase);
+
+    public ToolRegistration WithTool(
+        string name,
+        string description,
+        Func<JsonDocument, ValueTask<object?>> method
+    )
```
sd-st, line 125: I think we're largely going to want to do what we did for Java. I.e. let the user specify the parameters and return type in a idiomatic C# way, and also specify annotations similarly (probably with attributes?)

src/Anthropic/Tool/ToolRunnerExtensions.cs
```diff
@@ -0,0 +1,359 @@
...
+    public required string Name { get; init; }
+
+    public required string Description { get; init; }
+
+    public InputSchema GenerateSchema()
+    {
+        var options = JsonSerializerOptions.Default;
+        var jsonSchema = options.GetJsonSchemaAsNode(typeof(TArgument));
+
+        //TODO: ask tomer or stephen for a copy constructor in codegen for easy transformation
```
sd-st, line 209: I think you could be good; since InputSchema is a record class you should be able to just use the compiler-generated one (which works as intended for our purposes)

sd-st (commented):
Didn't review the full thing but left some comments

src/Anthropic/Anthropic.csproj
JPVenson: that was purely made temporary for easier review

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: i would argue its better to have that seperated because one is the data model and one is the actual logic and you should not have buissniss logic in your models. This also has the added benefit of keeping the custom code and the generated code seperate to avoid continues conflicts in the code

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: oh for sure, those are only temporary placeholders for actual anthropic exeptions

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: well close-ish. You are likely thinking of doing this:
```
await toolRunner?.Invoke(toolUseInputArguments).ConfigureAwait(false);
```
However null coaleasing operators dont work with async/await. (`toolRunner?.Invoke(toolUseInputArguments)` becomes null but you cannot await null)

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: umpf yes you are right i missread the java code there

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: that was another question i wanted to ask, the java code as well as the Ollama sharp example we talked about only use plaintext never structured results
https://github.com/anthropics/anthropic-sdk-java/blob/e3adc58c74d3dab38697bcd46f5845dcc390a458/anthropic-java-example/src/main/java/com/anthropic/example/BetaMessagesToolsRawExample.java#L86
https://github.com/awaescher/OllamaSharp/blob/main/src/OllamaSharp/Chat.cs#L469

for responses. I did build it to allow structured responses but could not find a way to provide them yet. Technically if the tooluseResult is an object or JsonDocument, it will be returned as stringified json atm.

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: Currently i build the base for two ways to do exactly that.

Either provide a (to be implemented) method or lambda which will then get read or provide an object that will be used.

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: `InputSchema` is one of the classes that do not accept an `JsonDocument` in its parameters and the internal _rawData is not accessable to me so i dont have any way setting that. Some objects do have a constructor that simply accepts a `JsonElement` or `JsonDocument` so if you could add that to codegen it would simply this code a lot

sd-st:
Also I just spoke to Sara and Java doesn't actually have a tool usage implementation, although the readme is still helpful (particularly the first section; it makes sense to me to expose tools the way they do, unless you can think of a better way)

TomerAberbach:
> Also I just spoke to Sara and Java doesn't actually have a tool usage implementation, although the readme is still helpful (particularly the first section; it makes sense to me to expose tools the way they do, unless you can think of a better way)

What do you mean by tool usage btw? Java's tool runner does support executing tools for you

sd-st:
> > Also I just spoke to Sara and Java doesn't actually have a tool usage implementation, although the readme is still helpful (particularly the first section; it makes sense to me to expose tools the way they do, unless you can think of a better way)
> 
> What do you mean by tool usage btw? Java's tool runner does support executing tools for you

Doesn't it require doing the tool invocation manually? That's what's in the readme: https://github.com/anthropics/anthropic-sdk-java/tree/main?tab=readme-ov-file#tool-use-with-json-schemas

TomerAberbach:
That's the manual way of doing it; I guess we don't have tool runner docs

Anyway, tool runner works this way: https://github.com/anthropics/anthropic-sdk-java/blob/main/anthropic-java-example/src/main/java/com/anthropic/example/BetaToolRunnerExample.java

And it'll call it for you

sd-st:
huh, seems like a bit of an oversight we don't document the tool runner in the readme. I imagine most people are just going to implement it manually since that's what we tell them to do

Am I correct the Java version is only implemented for the beta endpoint? If yes, any reason or is it just that that's all we got to? Looks like the non-beta endpoint may not support tools?

TomerAberbach:
> huh, seems like a bit of an oversight we don't document the tool runner in the readme. I imagine most people are just going to implement it manually since that's what we tell them to do

Yeah agreed

> Am I correct the Java version is only implemented for the beta endpoint? If yes, any reason or is it just that that's all we got to? Looks like the non-beta endpoint may not support tools?

I think it was only supported in beta, if I recall correctly? It was a while ago that I implemented the runner

src/Anthropic/Tool/ToolRunnerExtensions.cs
sd-st: Sure, that makes sense

src/Anthropic/Tool/ToolRunnerExtensions.cs
sd-st: This example may be more helpful: https://github.com/anthropics/anthropic-sdk-java/blob/main/anthropic-java-example/src/main/java/com/anthropic/example/BetaToolRunnerExample.java

If the Java version can only support string responses (I haven't checked) then this is probably fine. But in that case we probably shouldn't call `ToString` and should instead force the user to give us a string

src/Anthropic/Tool/ToolRunnerExtensions.cs
sd-st: In that case we probably want to make these methods internal - I don't see a good reason to let the user use them

src/Anthropic/Tool/ToolRunnerExtensions.cs
sd-st: Maybe I'm misunderstanding; what do you want exactly?

src/Anthropic/Tool/ToolRunnerExtensions.cs
JPVenson: hmm no that would defeat the purpose of having those methods entirely.

The goal of those methods is so that the user can do this:

```csharp

IMessageService service = ...;
MessageCreateParams arguments = ...;
service.CreateWithToolLookup(
	arguments, 
	tools => 
		tools.WithTool("get_weather", "gets the weather at location", toolArg => {...})
			 .WithTool<FooBaTool>()
	);

```

Thats what those extension methods are for. This would be the modern csharp idiomatic way of allowing subconfiguration for those scenarios.

sd-st: Alright sure

Issue #133, opened 2026-02-25
Support setting cache control for System messages when using AnthropicBetaClientExtensions 


1yefuwang1:
The implementation of the `AnthropicBetaClientExtensions` doesn't respect system messages' `RawRepresentations`, making it hard to apply caching for system messages.

https://github.com/anthropics/anthropic-sdk-csharp/blob/Anthropic-v12.8.0/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs#L519
```csharp
  private static List<BetaMessageParam> CreateMessageParams(
            IEnumerable<ChatMessage> messages,
            out List<BetaTextBlockParam>? systemMessages
        )
        {
            List<BetaMessageParam> messageParams = [];
            systemMessages = null;

            foreach (ChatMessage message in messages)
            {
                if (message.Role == ChatRole.System)
                {
                    foreach (AIContent content in message.Contents)
                    {
                        if (content is TextContent tc)
                        {
                            // Note, RawRepresentation is not honored. A new message is created.
                            (systemMessages ??= []).Add(new() { Text = tc.Text });
                        }
                    }

                    continue;
                }
..... other code omitted
}

```

sd-st:
@stephentoub Is this something you'd be interested in looking at?

sd-st:
This should be fixed now! Thanks for your report

PR #135, opened 2026-02-26
Preserve RawRepresentation in IChatClient system message mapping

stephentoub:
Fixes https://github.com/anthropics/anthropic-sdk-csharp/issues/133

cc: @sd-st

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
+++ b/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
@@ -1628,6 +1628,71 @@ public class AnthropicClientBetaExtensionsTests : AnthropicClientExtensionsTests
     }
 
     [Fact]
+    public async Task GetResponseAsync_SystemMessageRawRepresentationPreserved()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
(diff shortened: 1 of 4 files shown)


src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
```diff
@@ -1627,6 +1627,71 @@ public async Task GetResponseAsync_WithRawRepresentationFactory_SystemMessagesLi
...
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            [
+                new ChatMessage(ChatRole.System, [systemContent]),
+                new ChatMessage(ChatRole.User, "Test"),
+            ],
+            new(),
+            TestContext.Current.CancellationToken
+        );
+        Assert.NotNull(response);
```
sd-st, line 1692: Could we assert more here? Seems like these tests would pass even without the code changes unless I'm missing something

stephentoub: > Seems like these tests would pass even without the code changes unless I'm missing something

The assertions are happening in the verbatim handler, ensuring that the JSON that's sent by the client is what's expected. Without the fixes in this PR, those will fail, because the RawRepresentation won't be propagated and this part of the expected input won't match:
```
                    "cache_control": {
                        "type": "ephemeral"
                    }
```

sd-st (dismissed):
Thanks!

sd-st:
Actually could you fix the file formatting?

stephentoub:
> Actually could you fix the file formatting?

@sd-st, done

sd-st:
Thanks!

PR #144, opened 2026-03-19
Fix IChatClient handling of empty messages when raw representation has messages

stephentoub:


src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -909,12 +904,23 @@ bool hasHostedFiles
             // or with only the required properties set.
             MessageCreateParams? createParams =
                 options?.RawRepresentationFactory?.Invoke(this) as MessageCreateParams;
+
+            // Anthropic requires at least one message. If no messages were provided either directly
+            // or via the RawRepresentationFactory, add a placeholder. It's currently expensive to access
+            // the Messages from the raw representation, as each access deserializes an entirely new collection,
+            // so we do it once upfront to try to reduce that overhead.
```
sd-st: FYI this is no longer true - there's a caching layer

sd-st (commented):
I would probably just change the comments and then this LGTM

sd-st (approved):
Thanks!

PR #147, opened 2026-03-19
fix: handle oversized SSE events in Bedrock SseEventContentWrapper

### LINKED ISSUE #146: Anthropic.Bedrock: SseEventContentWrapper.ReadAsync throws ArgumentException when SSE event exceeds buffer size
## Description

`SseEventContentWrapper.SseLazyEventStream.ReadAsync` throws `System.ArgumentException: Destination is too short` when an AWS EventStream message, after being converted to SSE format, is larger than the caller's read buffer.

## Location

`src/Anthropic.Bedrock/SseEventContentWrapper.cs`, line 80:

```csharp
var encodedData = Encoding.UTF8.GetBytes(data!);
encodedData.CopyTo(buffer);  // ← No bounds check
return encodedData.Length;
```

`CopyTo(Memory<byte>)` throws if `encodedData.Length > buffer.Length`.

## Reproduction

Use `AnthropicBedrockClient` with streaming (`CreateStreaming`) on responses that produce large SSE events (e.g., long text blocks, tool call results). The `SseParser` provides a fixed-size buffer, and when a single eventstream message produces an SSE payload larger than that buffer, the copy fails.

## Stack trace

```
System.ArgumentException: Destination is too short. (Parameter 'destination')
   at Anthropic.Bedrock.SseEventContentWrapper+SseLazyEventStream.ReadAsync(Memory<byte>, CancellationToken)
   at System.Net.ServerSentEvents.SseParser<T>.FillLineBufferAsync(CancellationToken)
   at System.Net.ServerSentEvents.SseParser<T>.EnumerateAsync(CancellationToken)+MoveNext()
   at Anthropic.Core.Sse.Enumerate<T>(HttpResponseMessage, CancellationToken)+MoveNext()
```

## Suggested fix

The `ReadAsync` implementation should respect the buffer size and return partial data, letting the caller call again for the remainder — standard `Stream.ReadAsync` contract:

```csharp
public override async ValueTask<int> ReadAsync(
    Memory<byte> buffer, CancellationToken cancellationToken = default)
{
    if (_remainder.Length > 0)
    {
        var toCopy = Math.Min(_remainder.Length, buffer.Length);
        _remainder.Slice(0, toCopy).CopyTo(buffer);
        _remainder = _remainder.Slice(toCopy);
        return toCopy;
    }

    var (data, success) = await AwsEventStreamHelpers
        .ReadStreamMessage(_sourceStream, cancellationToken)
        .ConfigureAwait(false);
    if (!success) return 0;

    var encoded = Encoding.UTF8.GetBytes(data!);
    var bytesToCopy = Math.Min(encoded.Length, buffer.Length);
    encoded.AsMemory(0, bytesToCopy).CopyTo(buffer);
    _remainder = encoded.AsMemory(bytesToCopy);
    return bytesToCopy;
}
```

## Environment

- `Anthropic.Bedrock` 0.1.0
- `Anthropic` 12.9.0
- .NET 10
- AWS Bedrock with cross-region inference profiles (`us.anthropic.claude-opus-4-6-v1`)

scottt732:
## Summary

- Fixes `ArgumentException: Destination is too short` in `SseEventContentWrapper.SseLazyEventStream.ReadAsync` when an AWS EventStream message exceeds the caller's read buffer
- Buffers remainder across subsequent `ReadAsync` calls per the standard `Stream.ReadAsync` contract
- Fixes both `#if NET` (Memory) and `#else` (byte array) code paths

Fixes #146


src/Anthropic.Bedrock/Anthropic.Bedrock.csproj
```diff
@@ -7,9 +7,13 @@
     <LangVersion>13.0</LangVersion>
     
     <PackageId>Anthropic.Bedrock</PackageId>
-    <VersionPrefix>0.1.0</VersionPrefix>
```
sd-st, line 10: FYI this gets bumped automatically; we shouldn't update it ourselves

sd-st (commented):
Thanks for your PR! If you could address the comment and also target `next` then this will be perfect (:

sd-st:
Oh and also run a `./scripts/format` to fix the formatting, otherwise CI will fail

scottt732:
all set! thanks

sd-st:
Looks like the formatting is still messed up; the line endings don't match. I fixed locally but looks like I don't have permission to push to your branch

scottt732:
sorry about that... had core.autocrlf issue.

sd-st (approved):
Thanks so much!

PR #152, opened 2026-03-27
fix(client): update Bearer scheme casing to match AWS requirement

### LINKED ISSUE #151: Bearer capitalisation in AnthropicBedrockApiTokenCredentials causes errors
This really isn't the fault of the anthropic library.. 

https://auth0.com/blog/the-bearer-token-case/

There's endless discussion around what the spec for bearer, er, Bearer auth case sensitivity really means, but in the interests of getting things done, I think the anthropic Bedrock client should just use "Bearer" as that's what actually works with AWS.

As it currently is, it causes other errors where it's looking for other AWS auth headers.  The fix is pretty simple, just "bearer" to "Bearer" on this line:

https://github.com/anthropics/anthropic-sdk-csharp/blob/d85867062e81edad41830c5f2522d018c896a834/src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs#L30

Have tested locally and it fixes the issue, will make a PR for it soon.  For anyone waiting, you can just add a class like this:

```
/// <summary>
/// Straight copy/paste of AnthropicBedrockApiTokenCredentials that addresses bearer auth issue
/// </summary>
public sealed class TotallyBearableAnthropicBedrockApiTokenCredentials : IAnthropicBedrockCredentials
{
    public required string BearerToken { get; init; }
    public required string Region { get; init; }

    public Task Apply(HttpRequestMessage requestMessage)
    {
        //AWS falls into the 'no case insensitivity is allowed' camp, so we need to use 'Bearer' in all it's case sensitive glory.
        //No judgement here.  OK, maybe just a little.
        requestMessage.Headers.Authorization =
            new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", BearerToken);
        return Task.CompletedTask;
    }
}
```

mstancombe:
Fixes #151

### DIFF
--- a/src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
+++ b/src/Anthropic.Bedrock/AnthropicBedrockApiTokenCredentials.cs
@@ -27,7 +27,7 @@ public sealed class AnthropicBedrockApiTokenCredentials : IAnthropicBedrockCrede
     public Task Apply(HttpRequestMessage requestMessage)
     {
         requestMessage.Headers.Authorization =
-            new System.Net.Http.Headers.AuthenticationHeaderValue("bearer", BearerToken);
+            new System.Net.Http.Headers.AuthenticationHeaderValue("Bearer", BearerToken);
         return Task.CompletedTask;
     }
 }


TomerAberbach (approved):

sd-st:
@mstancombe could you rebase this to point at the `next` branch?

mstancombe:
@TomerAberbach sorry, only just saw that rebase req, tks for sorting it.

PR #154, opened 2026-03-29
fix: make ToolUseBlock.Caller nullable for non-MCP tool calls

kibo-thomasphipps:
## Summary

- `ToolUseBlock.Caller` was marked `required` with `GetNotNullClass`, but the API does not include a `caller` field in `tool_use` content blocks for standard (non-MCP) tool calls. This caused `AnthropicInvalidDataException` ("'caller' cannot be absent") during streaming deserialization.
- Changed `Caller` from `required ToolUseBlockCaller` to `ToolUseBlockCaller?`, using `GetNullableClass` and a null-guarding init setter consistent with other optional properties in the codebase.
- Added tests covering unset and explicitly-null `Caller` for both property access and validation.

## Reproduction

```
Anthropic.Exceptions.AnthropicInvalidDataException: 'caller' cannot be absent
   at Anthropic.Core.JsonDictionary.GetNotNullClass[T](String key)
   at Anthropic.Models.Messages.ToolUseBlock.Caller.get()
   at Anthropic.Models.Messages.ToolUseBlock.Validate()
   at Anthropic.Models.Messages.RawContentBlockStartEvent.ContentBlock.get()
```

Any Claude model returning a standard `tool_use` block (non-MCP) triggers this during streaming via `AsIChatClient()`.

## Test plan

- [x] `OptionalNonNullablePropertiesUnsetAreNotSet_Works` — `Caller` is null and absent from `RawData` when not set
- [x] `OptionalNonNullablePropertiesUnsetValidation_Works` — `Validate()` passes without `Caller`
- [x] `OptionalNonNullablePropertiesSetToNullAreNotSet_Works` — explicit `null` does not insert into `RawData`
- [x] `OptionalNonNullablePropertiesSetToNullValidation_Works` — `Validate()` passes with explicit `null`
- [x] Existing `ToolUseBlockTest` and `ToolUseBlockCallerTest` tests continue to pass

### DIFF
--- a/src/Anthropic.Tests/Models/Messages/ToolUseBlockTest.cs
+++ b/src/Anthropic.Tests/Models/Messages/ToolUseBlockTest.cs
@@ -144,6 +144,78 @@ public class ToolUseBlockTest : TestBase
 
         Assert.Equal(model, copied);
     }
+
+    [Fact]
+    public void OptionalNonNullablePropertiesUnsetAreNotSet_Works()
+    {
+        var model = new ToolUseBlock
+        {
+            ID = "id",
+            Input = new Dictionary<string, JsonElement>()
+            {
+                { "foo", JsonSerializer.SerializeToElement("bar") },
+            },
+            Name = "x",
+        };
+
+        Assert.Null(model.Caller);
+        Assert.False(model.RawData.ContainsKey("caller"));
+    }
+
+    [Fact]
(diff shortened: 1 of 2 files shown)


sd-st:
Could you open an issue an explain a bit more about what the problem is, ideally with some example code?

We generally fix things like this by updating the spec and then regenerating the SDK rather than updating the SDK directly.

Issue #150, opened 2026-03-26
CodeExecutionTool in Microsoft.Extensions.Ai doesn't return ToolResults


flintlouis:
Using "Anthropic" Version="12.9.0"
When using Microsoft.Extensions.Ai, adding the CodeExecutionTool, I can't seem to get the results of that tool when using the Microsoft.Extensions.Ai AiContent CodeInterpreterToolResultContent and CodeInterpreterToolCallContent results?

Will this be integrated later? Am I missing a header? Am I dong something wrong?
This is my code snippit:
```
List<ToolUnion> tools =
[
	new(new CodeExecutionTool20250825())
];

options.RawRepresentationFactory = _ => new MessageCreateParams
{
	MaxTokens = 4_069,
	Messages = [],
	Model = config.Model,
	Tools = tools,
};
```

flintlouis:
The same is true for WebSearch tool use. The WebSearchToolResultContent is not filled

sd-st:
@stephentoub mind taking a look at this?

stephentoub:
@sd-st https://github.com/anthropics/anthropic-sdk-csharp/pull/118 moved that support from the Beta to the stable implementation. As far as I can tell, no new 12.10 nuget package has shipped since then, so this functionality hasn't been released?

sd-st:
We just had a release go out today; @flintlouis can you confirm that this is still an issue?

flintlouis:
@sd-st I updated Anthropic to version 12.10.0 and tested a little bit. It seems that if you use `ChatResponseresponse = await chatClient.GetResponseAsync(messages, options);` it works. `CodeInterpreterToolCallContent` and `CodeInterpreterToolResultContent` are both set and Inputs and Outputs are both populated.
But when using `IAsyncEnumerable<ChatResponseUpdate> responseStream = chatClient.GetStreamingResponseAsync(messages, options)` the `CodeInterpreterToolCallContent.Inputs` is not populated.

Here is my code setup:
```
List<ChatMessage> messages = [
	new ChatMessage(ChatRole.System, "You are a personal math tutor. When asked a math question, write and run code to answer the question."),
	new ChatMessage(ChatRole.User, "Calculate the mean and standard deviation of [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]")
];

List<ChatResponseUpdate> responseUpdates = [];

[...]

PR #162, opened 2026-04-03
Fix AIContent creation for streaming server tool output

stephentoub:
https://github.com/anthropics/anthropic-sdk-csharp/issues/150#issuecomment-4168669772


sd-st (approved):
Thanks!

PR #166, opened 2026-04-06
fix(client): merge response format into preconfigured output config

### LINKED ISSUE #165: IChatClient adapter skips ResponseFormat when OutputConfig is pre-set via RawRepresentationFactory (Effort + JSON structured output conflict)
### Description

The `AnthropicChatClient` adapter does not merge `OutputConfig` properties. When `OutputConfig` is pre-set via `RawRepresentationFactory` (e.g., to configure `Effort`), the adapter skips mapping `ChatOptions.ResponseFormat` to `OutputConfig.Format`, resulting in structured JSON output being **silently dropped**.

---

### Root Cause

In [`AnthropicClientExtensions.cs#L1186`](https://github.com/anthropics/anthropic-sdk-csharp/blob/0bef96bd/src/Anthropic/AnthropicClientExtensions.cs#L1186):

```csharp
if (createParams.OutputConfig is null && options.ResponseFormat is { } responseFormat)
```

The guard is a binary `is null` check. If `OutputConfig` is already set (e.g., with only `Effort`), **the entire block is skipped** — including the `Format` assignment. There is no merging of individual `OutputConfig` properties.

---

### Reproduction

```csharp
var client = new AnthropicClient { ApiKey = "..." };
IChatClient chatClient = client.AsIChatClient("claude-opus-4-6");

var options = new ChatOptions
{
    ResponseFormat = ChatResponseFormat.ForJsonSchema<MyResponseType>(),
    RawRepresentationFactory = _ => new MessageCreateParams
    {
        Model = Model.ClaudeOpus4_6,
        MaxTokens = 16384,
        Messages = [],
        OutputConfig = new OutputConfig
        {
            Effort = Effort.Medium // ⚠️ Setting Effort causes Format to be silently skipped
        }
    }
};

// BUG: Request is sent WITHOUT the JSON schema format.
// Model returns free-form text instead of structured JSON.
// With low MaxTokens, response is truncated with output_tokens: 0 or 1.
```

---

### Expected Behavior

The adapter should merge `OutputConfig.Format` from `ChatOptions.ResponseFormat` even when `OutputConfig` is already set. Both `Effort` and `Format` are independent properties on `OutputConfig` and should be settable together.

---

### Proposed Fix

**Change line 1186 from:**

```csharp
if (createParams.OutputConfig is null && options.ResponseFormat is { } responseFormat)
```

**To:**

```csharp
if (createParams.OutputConfig?.Format is null && options.ResponseFormat is { } responseFormat)
```

**And update the assignment to preserve existing `OutputConfig` properties:**

```csharp
createParams = createParams with
{
    OutputConfig = (createParams.OutputConfig ?? new OutputConfig()) with
    {
        Format = new JsonOutputFormat() { ... }
    }
};
```

---

### Additional Context

This also affects **`Model`** and **`MaxTokens`** — when `RawRepresentationFactory` provides `MessageCreateParams`, the adapter does not override `Model` from `ChatOptions.ModelId` or `MaxTokens` from `ChatOptions.MaxOutputTokens`. This forces consumers to duplicate these values in the factory, defeating the purpose of the abstraction.

Related issue: #164 (similar pattern — `AdditionalProperties` not mapped for `CacheControl`)

---

### Impact

Without this fix, it is **impossible** to use both `Effort` and structured JSON output (`ChatOptions.ResponseFormat`) through the `IChatClient` abstraction. The only workaround is to duplicate the adapter's JSON schema construction logic inside `RawRepresentationFactory`, which is fragile and tightly coupled to SDK internals.

---

### Environment

| | |
|---|---|
| **Anthropic SDK** | 12.10.0 |
| **Microsoft.Extensions.AI** | 10.4.1 |
| **.NET** | 10 |

jozkee:
## Summary
- merge ChatOptions.ResponseFormat into an existing OutputConfig instead of skipping it when RawRepresentationFactory preconfigures output settings
- keep beta behavior aligned by only checking the canonical OutputConfig.Format
- add regression coverage for standard and beta chat adapters

Fixes #165.

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
+++ b/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
@@ -912,6 +912,91 @@ public class AnthropicClientBetaExtensionsTests : AnthropicClientExtensionsTests
     }
 
     [Fact]
+    public async Task GetResponseAsync_WithResponseFormatAndPreconfiguredOutputConfig_MergesFormat()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 2048,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Return JSON"
+                    }]
+                }],
+                "output_config": {
+                    "effort": "medium",
+                    "format": {
+                        "type": "json_schema",
+                        "schema": {
+                            "type": "object",
+                            "properties": {
+                                "answer": { "type": "string" }
+                            },
+                            "required": ["answer"],
+                            "additionalProperties": false
+                        }
+                    }
+                }
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_factory_output_config_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "{\"answer\":\"Response\"}"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 20,
+                    "output_tokens": 5
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        ChatOptions options = new()
+        {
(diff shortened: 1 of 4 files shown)


sd-st:
Looks like this needs to be formatted; mind running a `./scripts/format`?

sd-st (approved):
Thanks! Just needs that format and then this is good to merge

PR #173, opened 2026-04-10
fix(client): Treat empty json string in meai tool call as having 'no' arguments.

mstancombe:
If a non-server tool has no arguments, it's currently still run through the json deserializer, which raises a JsonException.  

This doesn't stop the chat client from working, CreateFromParsedArguments just seems to log the exception in the created FunctionCallContent, but it does mean a noisy exception, and the created FunctionCallContent doesn't reflect the true state of play.

This PR handles an empty string as meaning there was no arguments given, which I think is how we should handle this.

### DIFF
--- a/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
+++ b/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
@@ -1921,13 +1921,15 @@ public static class AnthropicBetaClientExtensions
                     functionData.CallId,
                     functionData.Name,
                     json =>
-                        (Dictionary<string, object?>?)
-                            JsonSerializer.Deserialize(
-                                json,
-                                AIJsonUtilities.DefaultOptions.GetTypeInfo(
-                                    typeof(Dictionary<string, object?>)
+                        json.Length == 0
+                            ? null
+                            : (Dictionary<string, object?>?)
+                                JsonSerializer.Deserialize(
+                                    json,
+                                    AIJsonUtilities.DefaultOptions.GetTypeInfo(
+                                        typeof(Dictionary<string, object?>)
+                                    )
                                 )
-                            )
                 );
             }
 
--- a/src/Anthropic/AnthropicClientExtensions.cs
+++ b/src/Anthropic/AnthropicClientExtensions.cs
@@ -1996,13 +1996,15 @@ public static class AnthropicClientExtensions
                     functionData.CallId,
                     functionData.Name,
                     json =>
-                        (Dictionary<string, object?>?)
-                            JsonSerializer.Deserialize(
-                                json,
-                                AIJsonUtilities.DefaultOptions.GetTypeInfo(
-                                    typeof(Dictionary<string, object?>)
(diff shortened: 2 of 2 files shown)


sd-st:
Would you be able to rebase this?

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1831,8 +1831,9 @@ private static AIContent CreateStreamingToolCallContent(StreamingFunctionData fu
                     functionData.Arguments.ToString(),
                     functionData.CallId,
                     functionData.Name,
-                    json =>
-                        (Dictionary<string, object?>?)
+                    json => json.Length == 0 
+                        ? null
```
sd-st: @stephentoub Does `null` make sense here or should re be returning an empty dictionary?

mstancombe:
Rebased to latest next.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
mstancombe: Just to expand on my reasoning, as I did look into this a bit before deciding on null, the 'Arguments' on ```FunctionCallContext``` class in MEAI looks like this:

```
/// <summary>
/// Gets or sets the arguments requested to be provided to the function.
/// </summary>
public IDictionary<string, object?>? Arguments { get; set; }
```

Which doesn't explicitly state *when* this should be null, but given that it *is* nullable, I think an llm passing no arguments probably should be interpreted as nothing.

Additionally, the debug view for ```FunctionCallContext``` does this:
```
/// <summary>Gets a string representing this instance to display in the debugger.</summary>
[DebuggerBrowsable(DebuggerBrowsableState.Never)]
private string DebuggerDisplay
{
    get
    {
        string display = "FunctionCall = ";

        if (CallId is not null)
        {
            display += $"{CallId}, ";
        }

        display += Arguments is not null ?
            $"{Name}({string.Join(", ", Arguments)})" :
            $"{Name}()";

        return display;
    }
}
```
If arguments was an empty dict, this would show as 'MyTool()' vs 'MyTool' when it's null.  Again, this doesn't prove what the intention was, but I'm guessing (hoping :P) that @stephentoub was expecting a null to represent 'no args'.

mstancombe: ah, disregard the final point, it would still show MyTool() with no args.. it just would skip the string.join over the empty dictionary, but no output difference.

mstancombe:
rebased to latest next again.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
sd-st: alright sure, this makes sense to me. I guess null is also slightly cheaper than instantiating an empty dictionary.

sd-st (approved):

sd-st:
Thanks!

PR #175, opened 2026-04-10
Make ServerToolUsage/BetaServerToolUsage web request counts nullable

### LINKED ISSUE #174: ServerToolUsage/BetaServerToolUsage: make web_*_requests nullable when key is absent
## Summary
`BetaServerToolUsage` and `ServerToolUsage` currently model `web_fetch_requests` / `web_search_requests` as `required long` and read them via `GetNotNullStruct<long>()`.

In real responses, these fields can be absent. When absent, property access (and `Validate()`) throws `AnthropicInvalidDataException`.

## Affected files
- `src/Anthropic/Models/Beta/Messages/BetaServerToolUsage.cs`
- `src/Anthropic/Models/Messages/ServerToolUsage.cs`

## Current behavior
If raw payload omits one of these keys, accessing the property throws due to non-null required getter.

## Expected behavior
Missing keys should be treated as optional/null, not as invalid payload.

## Suggested fix
1. Remove `required` from both properties in both classes.
2. Change type from `long` to `long?`.
3. Use `GetNullableStruct<long>(...)` instead of `GetNotNullStruct<long>(...)`.
4. Adjust `Validate()` so absent values do not force a throw.
5. Add tests for missing-key scenarios in both beta and non-beta test suites.

## Why this helps
This aligns the model with observed server payload shape and avoids runtime exceptions for legitimate responses.

Sunny-DotNet:
## Summary
- make WebFetchRequests and WebSearchRequests nullable (long?) in both ServerToolUsage and BetaServerToolUsage
- switch getters to GetNullableStruct<long>(...)
- update Validate() to validate keys only when present
- update usage-mapping callsites to handle nullable values safely
- add missing-field tests for both beta and non-beta models

## Why
web_fetch_requests and web_search_requests can be absent in real payloads. Treating them as required caused runtime exceptions.

Fixes #174

### DIFF
--- a/src/Anthropic.Tests/Models/Beta/Messages/BetaServerToolUsageTest.cs
+++ b/src/Anthropic.Tests/Models/Beta/Messages/BetaServerToolUsageTest.cs
@@ -11,8 +11,8 @@ public class BetaServerToolUsageTest : TestBase
     {
         var model = new BetaServerToolUsage { WebFetchRequests = 2, WebSearchRequests = 0 };
 
-        long expectedWebFetchRequests = 2;
-        long expectedWebSearchRequests = 0;
+        long? expectedWebFetchRequests = 2;
+        long? expectedWebSearchRequests = 0;
 
         Assert.Equal(expectedWebFetchRequests, model.WebFetchRequests);
         Assert.Equal(expectedWebSearchRequests, model.WebSearchRequests);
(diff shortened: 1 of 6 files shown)


Sunny-DotNet:
Closing this PR in favor of a replacement PR with the same fix on a fresh branch.

PR #180, opened 2026-04-18
Handle missing streaming thinking signatures

xoofx:
## Summary
- stop the Microsoft.Extensions.AI Anthropic adapters from throwing when a streaming thinking block omits the initial `signature`
- read `signature` from raw JSON opportunistically for thinking blocks in both the standard and beta adapters
- add regression coverage for providers that emit the signature later via `signature_delta`

## Why
MiniMax''s Anthropic-compatible streaming API sends a `thinking` content block without an initial `signature`, then emits the signature later as a `signature_delta`. The current adapters access `thinking.Signature` eagerly, which throws before the stream can continue.

This change keeps the existing `signature_delta` handling, but avoids assuming the initial block always includes a signature.

## Testing
- `dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj -c Release --no-restore --filter "GetStreamingResponseAsync_WithThinkingBlockMissingInitialSignature_DoesNotThrow|GetStreamingResponseAsync_WithThinkingBlockInStream|GetStreamingResponseAsync_WithSignatureDeltaInStream"`

### DIFF
--- a/src/Anthropic/Models/Beta/Messages/BetaThinkingBlock.cs
+++ b/src/Anthropic/Models/Beta/Messages/BetaThinkingBlock.cs
@@ -16,7 +16,9 @@ public sealed record class BetaThinkingBlock : JsonModel
         get
         {
             this._rawData.Freeze();
-            return this._rawData.GetNotNullClass<string>("signature");
+            // Some APIs are advertised as Anthropic-compatible but erroneously omit the `signature` field in the response.
+            // This is a bug in the API itself, although we can handle it fairly easily here, so we do.
+            return this._rawData.GetNullableClass<string>("signature") ?? string.Empty;
         }
         init { this._rawData.Set("signature", value); }
     }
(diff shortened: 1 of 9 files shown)


xoofx:
Oops an agent created this PR (it's an actual issue I got) prematurely but I haven't checked the code before it created the PR. 😅 Will double check locally.

xoofx:
I have corrected the fix and used a `string.Empty` instead.

sd-st:
> MiniMax's Anthropic-compatible streaming API sends a thinking content block without an initial signature, then emits the signature later as a signature_delta. The current adapters access thinking.Signature eagerly, which throws before the stream can continue.

FWIW I would not call it Anthropic-compatible if it doesn't work with the unmodified SDK :sweat_smile: This problem will still exist in other SDKs but I'm okay with this PR conceptually. It would be better if this bug was fixed in the API

src/Anthropic/Models/Beta/Messages/BetaThinkingBlock.cs
```diff
@@ -16,7 +16,7 @@ public required string Signature
         get
         {
             this._rawData.Freeze();
-            return this._rawData.GetNotNullClass<string>("signature");
+            return this._rawData.GetNullableClass<string>("signature") ?? string.Empty; // Can be null on some Anthropic "compatible" API responses
```
sd-st: I would expand this out a bit, e.g.


```suggestion
            // Some APIs are advertised as Anthropic-compatible but erroneously omit the `signature` field in the response.
            // This is a bug in the API itself, although we can handle it fairly easily here, so we do.
            return this._rawData.GetNullableClass<string>("signature") ?? string.Empty;
```

And then can also do the same thing for the other file

sd-st:
Can you fix the failing CI and also retarget this to `next`?

src/Anthropic/Models/Beta/Messages/BetaThinkingBlock.cs
xoofx: Fíxed via 2b57627

xoofx:
Going to reopen and target next

PR #181, opened 2026-04-22
Handle missing streaming thinking signatures

xoofx:
Followup from #180 

## Summary
- stop the Microsoft.Extensions.AI Anthropic adapters from throwing when a streaming thinking block omits the initial `signature`
- read `signature` from raw JSON opportunistically for thinking blocks in both the standard and beta adapters
- add regression coverage for providers that emit the signature later via `signature_delta`

## Why
MiniMax''s Anthropic-compatible streaming API sends a `thinking` content block without an initial `signature`, then emits the signature later as a `signature_delta`. The current adapters access `thinking.Signature` eagerly, which throws before the stream can continue.

This change keeps the existing `signature_delta` handling, but avoids assuming the initial block always includes a signature.

## Testing
- `dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj -c Release --no-restore --filter "GetStreamingResponseAsync_WithThinkingBlockMissingInitialSignature_DoesNotThrow|GetStreamingResponseAsync_WithThinkingBlockInStream|GetStreamingResponseAsync_WithSignatureDeltaInStream"`

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -4090,6 +4090,92 @@ public abstract class AnthropicClientExtensionsTestsBase
     }
 
     [Fact]
+    public async Task GetStreamingResponseAsync_WithThinkingBlockMissingInitialSignature_DoesNotThrow()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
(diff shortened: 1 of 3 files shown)


sd-st:
@xoofx Would you be able to rebase this? Looks like some unrelated changes made it in

xoofx:
> @xoofx Would you be able to rebase this? Looks like some unrelated changes made it in

done

sd-st (approved):

sd-st:
@xoofx just needs the failing test fixed and this should be good to go

sd-st:
Thank you!

PR #182, opened 2026-04-22
Fix several HasNext() that should rely on response.HasMore

xoofx:
When trying the API on Kimi 2.6 provider, I discovered that the `ModelListPage.cs` with HasNext() was looping forever (and eventually failed).

Their LastId is not null and always equal to the FirstId. While the `HasMore` of the response is correctly set.

This PR is addressing this by replacing all HasNext() for which we can use a `response.HasMore` instead.


sd-st:
Is this reproducible on Anthropic's API?

xoofx:
> Is this reproducible on Anthropic's API?

I don't think so because the .NET SDK was built with the behavior of Anthropic endpoint.

I just tested it with an Anthropic endpoint and `has_more` seems to be correctly reported.

As `has_more` is part of the API and is explicitly stated to be used for pagination. I don't know why the previous pattern `return this.Items.Count > 0 && response.LastID != null;` was used, but it can break with other providers (e.g. LastId is never null and == FirstId, to it loops)

xoofx:
I have rebased on latest `next` branch.

xoofx:
Hello, would it be possible to merge this PR? It is blocking the usage of Anthropic API with other model providers. I have to maintain a fork of this project and add it to my projects locally in order to get this fix.

sd-st:
@xoofx Hey sorry, just bringing this up with the team internally given that this only affects a 3rd party API.

xoofx:
> @xoofx Hey sorry, just bringing this up with the team internally given that this only affects a 3rd party API.

Ok, thanks. In case this is not accepted, I would like to understand why the existing code is not using `has_more` (the rationale, was it not working before? was it introduced after?). If really that doesn't make it that's ok, I will fork the project in my own project, annoying but workaround-able. 🙂

xoofx:
Hello @sd-st, have you been able to settle on a decision with the team?
Also @stephentoub if you have any comments?

sd-st (approved):
Tested and working. I spoke with the team and while we'd like to backport this to codegen, it's pretty low priority, so I'm happy to stick with this change for now. Thanks!

xoofx:
Great, thank you!

Issue #169, opened 2026-04-08
BetaImageBlockParam(new BetaFileImageSource wont add Beta headers


rodriciru:
```csharp
userMessage.Contents.Add(new Microsoft.Extensions.AI.AIContent() { RawRepresentation = new BetaContentBlockParam(new BetaImageBlockParam(new BetaFileImageSource(uploadFile.ID))) });
```
This code won't add the neccesary beta headers, so Claude complains `messages.0.content.1.image.source: Input tag 'file' found using 'type' does not match any of the expected tags: 'base64', 'url'`

Workarround:
```csharp
 options.RawRepresentationFactory = _ => new MessageCreateParams
{
    Model = model,
    MaxTokens = 1024,
    Messages = [],
    Betas = [Anthropic.Models.Beta.AnthropicBeta.FilesApi2025_04_14],
};
```

sd-st:
@stephentoub Anything we could do to improve the experience here? Not sure if we want to add the headers automatically or add some kind of mechanism to make it easier to add the betas

FWIW in the "normal" flow (i.e. no `Microsoft.Extensions.AI`) we don't add the extension automatically

stephentoub:
> Anything we could do to improve the experience here?

We could add the headers automatically; my assumption has been that we don't want to, that Anthropic is very explicitly choosing to make it opt-in.

But if adding the headers is desirable, it's reasonably easy to do.

sd-st:
If there's a good way to make them relatively easy to opt into (like when hitting the vanilla API) that would probably be best for now. We do populate the headers automatically for some params, just not `MessageCreateParams`. It's probably better to be consistent

stephentoub:
> If there's a good way to make them relatively easy to opt into

The easiest way would be fixing MessageCreateParams' Model, MaxTokens, and Messages to not be required.

Then instead of:
```csharp
options.RawRepresentationFactory = _ => new MessageCreateParams
{
    Model = model,
    MaxTokens = 1024,
    Messages = [],
    Betas = [Anthropic.Models.Beta.AnthropicBeta.FilesApi2025_04_14],
};
```
it'd just be:
```csharp
options.RawRepresentationFactory = _ => new MessageCreateParams
{
    Betas = [Anthropic.Models.Beta.AnthropicBeta.FilesApi2025_04_14],
};
```
which is pretty minimal.

stephentoub:
[...]

PR #183, opened 2026-04-23
fix(client): map HostedFileContent with image media type to image block

jozkee:
When HostedFileContent has an image MediaType (e.g. image/png), map it to `BetaImageBlockParam` instead of always using `BetaRequestDocumentBlock` to allows users to send hosted file images via MEAI abstractions without dropping down to `RawRepresentation` (break glass).

Applied in both user message content conversion and tool result content conversion paths.

Also fix another bug where we were not sending the beta header when the `HostedFileContent` was in a `FunctionResultContent`.

### DIFF
--- a/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
+++ b/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
@@ -232,12 +232,9 @@ public static class AnthropicBetaClientExtensions
 
             List<BetaMessageParam> messageParams = CreateMessageParams(
                 messages,
-                out List<BetaTextBlockParam>? systemMessages
+                out List<BetaTextBlockParam>? systemMessages,
+                out bool hasHostedFiles
             );
-            bool hasHostedFiles = messages
-                .SelectMany(m => m.Contents)
-                .OfType<HostedFileContent>()
-                .Any();
             MessageCreateParams createParams = GetMessageCreateParams(
                 messageParams,
                 systemMessages,
@@ -304,12 +301,9 @@ public static class AnthropicBetaClientExtensions
 
             List<BetaMessageParam> messageParams = CreateMessageParams(
                 messages,
-                out List<BetaTextBlockParam>? systemMessages
+                out List<BetaTextBlockParam>? systemMessages,
+                out bool hasHostedFiles
             );
-            bool hasHostedFiles = messages
-                .SelectMany(m => m.Contents)
-                .OfType<HostedFileContent>()
-                .Any();
             MessageCreateParams createParams = GetMessageCreateParams(
                 messageParams,
                 systemMessages,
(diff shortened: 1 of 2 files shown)


jozkee:
@sd-st re: [this comment](https://github.com/anthropics/anthropic-sdk-csharp/issues/169#issuecomment-4281165388), that would set a precedent of inspecting `RawRepresentation` to derive headers. I think it's better to extend first-class support so customers don't need to break glass in the first place.

jeffhandley (approved):
Reviewed from the MEAI perspective; looks good to me.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -805,9 +827,12 @@ DataContent dc when dc.HasTopLevelMediaType("text") =>
...
                                                         ),
                                                     }
-                                                ),
+                                                )
+                                            );
+                                            break;
 
-                                            UriContent uc when uc.HasTopLevelMediaType("image") =>
+                                        case UriContent uc when uc.HasTopLevelMediaType("image"):
+                                            blocks.Add(
```
sd-st, line 835: Should we have a `hasHostedFiles = true;` in this block?

sd-st (changes requested):

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
jozkee: Nope, image url contents don't need any headers. 
```sh
curl -sS -w "\nHTTP %{http_code}\n" https://api.anthropic.com/v1/messages \
   -H "x-api-key: $ANTHROPIC_API_KEY" \
   -H "anthropic-version: 2023-06-01" \
   -H "content-type: application/json" \
   -d '{
     "model": "claude-sonnet-4-5",
     "max_tokens": 256,
     "messages": [{
       "role": "user",
       "content": [
         {"type": "text", "text": "Describe the image in one short sentence."},
         {"type": "image", "source": {"type": "url", "url": "https://avatars.githubusercontent.com/u/76263028?s=200&v=4"}}
       ]
     }]
   }'
```
```
{"model":"claude-sonnet-4-5-20250929","id":"msg_01Y...","type":"message","role":"assistant","content":[{"type":"text","text":"A white \"AI\" logo appears on a dark background in a bold, modern typeface."}],"stop_reason":"end_turn","stop_sequence":null,"stop_details":null,"usage":{"input_tokens":83,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":22,"service_tier":"standard","inference_geo":"not_available"}}
HTTP 200
```

sd-st:
Sorry for not catching this earlier but could you rebase off of `next`? We generally only merge into main for releases. LGTM other than this

jozkee:
@sd-st done, can you please help me trigger another CI run?

sd-st:
Thanks!

PR #200, opened 2026-05-20
fix: streaming parallel tool-use emit logic (issue #197)

### LINKED ISSUE #197: GetStreamingResponseAsync emits duplicate/incomplete FunctionCallContent when multiple tools are called in parallel
## Bug Description

When the LLM calls multiple tools in parallel (e.g. ToolA + ToolB in the same response),
`GetStreamingResponseAsync` emits incorrect `FunctionCallContent` objects due to a logic
error in the `RawContentBlockStopEvent` handler.

## Root Cause

In `AnthropicClientExtensions.cs`, the `RawContentBlockStopEvent` handler iterates over
**all** entries in `streamingFunctions` instead of only the entry at `contentBlockStop.Index`:

```csharp
case RawContentBlockStopEvent contentBlockStop:
    if (streamingFunctions is not null)
    {
        foreach (var sf in streamingFunctions) // ← should only emit sf at contentBlockStop.Index
        {
            contents.Add(FunctionCallContent.CreateFromParsedArguments(...));
        }
        streamingFunctions.Clear();
    }
    break;
```

## What Actually Happens

Given a response with ToolA (index 1) and ToolB (index 2):

| Event | streamingFunctions | Emitted |
|---|---|---|
| ToolA `RawContentBlockStopEvent` | `{1: ToolA, 2: ToolB}` | FCC(ToolA) + FCC(ToolB) ← ToolB emitted early with incomplete Arguments |
| ToolB `RawContentBlockStopEvent` | `{}` (already cleared) | nothing ← ToolB's complete Arguments are lost |

This results in:
1. **ToolB is emitted before its `InputJsonDelta` chunks are complete** — Arguments may be incomplete
2. **`FunctionInvokingChatClient` receives duplicate/wrong `FunctionCallContent`**, causing it to
   send a second API request with duplicate `tool_use` ids, which Anthropic API rejects with:
{"type":"error","error":{"type":"invalid_request_error","message":"messages.X.content.Y: tool_use ids must be unique"}}
## Expected Behavior

`RawContentBlockStopEvent` should only emit the `FunctionCallContent` for the tool at
`contentBlockStop.Index`:

```csharp
case RawContentBlockStopEvent contentBlockStop:
    if (streamingFunctions is not null &&
        streamingFunctions.TryGetValue(contentBlockStop.Index, out var sf))
    {
        contents.Add(FunctionCallContent.CreateFromParsedArguments(
            sf.Arguments.ToString(),
            sf.CallId,
            sf.Name,
            json => (Dictionary<string, object?>?)JsonSerializer.Deserialize(
                json,
                AIJsonUtilities.DefaultOptions.GetTypeInfo(typeof(Dictionary<string, object?>))
            )
        ));
    }
    break;
```

## Steps to Reproduce

1. Use `AnthropicClient.AsIChatClient()` with `UseFunctionInvocation()` middleware
2. Provide a prompt that causes the LLM to call two or more tools in parallel in the same response turn
3. Observe `AnthropicBadRequestException` with `tool_use ids must be unique`

## Environment

- Package: `Anthropic` (official SDK)
- Affected versions: v11.0.0, confirmed still present in `main`
- `Microsoft.Extensions.AI` version: 10.4.1

jozkee:
RawContentBlockStopEvent was iterating over the entire streamingFunctions dictionary and clearing it. When a later tool_use block's content_block_start arrived before an earlier block's content_block_stop, the stop event would emit FunctionCallContent for every tracked tool, **including ones whose input_json deltas had not yet arrived**, and then wipe the dictionary, losing the later tools' real arguments.

Only emit (and remove) the entry whose index matches the stop event, in both AnthropicClientExtensions and AnthropicBetaClientExtensions. Add a regression test in AnthropicClientExtensionsTestsBase so coverage applies to both client variants.

Fixes #197

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -3061,6 +3061,87 @@ public abstract class AnthropicClientExtensionsTestsBase
     }
 
     [Fact]
+    public async Task GetStreamingResponseAsync_WithOverlappingParallelToolBlocks_EmitsOneCompleteFunctionCallPerBlock()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Call overlapping parallel tools"
+                    }]
+                }],
+                "stream": true
+            }
+            """,
+            actualResponse: """
+            event: message_start
+            data: {"type":"message_start","message":{"id":"msg_overlap_01","type":"message","role":"assistant","model":"claude-haiku-4-5","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":0}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"toolu_a","name":"tool_a","input":{},"caller":{"type":"direct"}}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":0,"delta":{"type":"input_json_delta","partial_json":"{\"arg\":\"a\"}"}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":1,"content_block":{"type":"tool_use","id":"toolu_b","name":"tool_b","input":{},"caller":{"type":"direct"}}}
+
+            event: content_block_stop
+            data: {"type":"content_block_stop","index":0}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"{\"arg\":\"b\"}"}}
+
+            event: content_block_stop
+            data: {"type":"content_block_stop","index":1}
+
+            event: message_delta
+            data: {"type":"message_delta","delta":{"stop_reason":"tool_use","stop_sequence":null},"usage":{"output_tokens":15}}
+
+            event: message_stop
+            data: {"type":"message_stop"}
+
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        List<ChatResponseUpdate> updates = [];
+        await foreach (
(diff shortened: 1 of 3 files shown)


src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -3060,6 +3060,87 @@ var update in chatClient.GetStreamingResponseAsync(
...
+            data: {"type":"message_start","message":{"id":"msg_overlap_01","type":"message","role":"assistant","model":"claude-haiku-4-5","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":0}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"toolu_a","name":"tool_a","input":{},"caller":{"type":"direct"}}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":0,"delta":{"type":"input_json_delta","partial_json":"{\"arg\":\"a\"}"}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":1,"content_block":{"type":"tool_use","id":"toolu_b","name":"tool_b","input":{},"caller":{"type":"direct"}}}
```
jozkee, line 3092: @ziv690922 what's still odd to me is how you could get `tool_use ids must be unique` if the ids are part of the content_block_start, they should never be duplicated, even in the buggy code.

ziv690922: You're right, the duplicate ids are not caused by this bug. After further investigation, the tool_use ids must be unique error was coming from FunctionInvokingChatClient internally appending the assistant turn (including tool_use blocks) into the messages list across iterations, which resulted in duplicate ids being sent to the Anthropic API. That's a separate issue from what this PR fixes.

TomerAberbach:
@stephentoub does this look right to you?

jeffhandley (approved):

stephentoub:
@TomerAberbach, sorry for the delay. This looks good to me. Also, for future reference, both @jozkee and @jeffhandley are on my team at Microsoft and I trust them with these kinds of fixes and improvements.

Suzii:
Hi, any update on this one? When could this fix be released?

sophie-ant (approved):
Thanks!

PR #202, opened 2026-05-22
Handle [DONE] terminal event in SSE streams from Azure APIM hosted Anthropic models

brused27:
Added a guard for `data: [DONE]` SSE terminal event emitted by Microsoft Foundry during SSE streams.  Without this guard, the SSE parser assigns a default `EventType` of `"message"` to such events, causing the SDK to attempt JSON deserialization on the literal string `"[DONE]"`. This results in a `JsonException` ('D' is an invalid start of a value).

Test case:
`dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj --filter "SseTest.Sse_SkipsDoneTerminalEvent"`

Previous exception seen with Anthropic model usage on Microsoft Foundry using streaming chat completion through Microsoft.Extensions.AI:
```
ERROR|Microsoft.Extensions.AI.LoggingChatClient|GetStreamingResponseAsync failed. Anthropic.Exceptions.AnthropicInvalidDataException: Message must be of type Anthropic.Models.Messages.RawMessageStreamEvent
 ---> System.Text.Json.JsonException: 'D' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
 ---> System.Text.Json.JsonReaderException: 'D' is an invalid start of a value. LineNumber: 0 | BytePositionInLine: 1.
   at System.Text.Json.ThrowHelper.ThrowJsonReaderException(Utf8JsonReader& json, ExceptionResource resource, Byte nextByte, ReadOnlySpan`1 bytes)
   at System.Text.Json.Utf8JsonReader.ConsumeValue(Byte marker)
   at System.Text.Json.Utf8JsonReader.ReadSingleSegment()
   at System.Text.Json.JsonSerializer.GetReaderScopedToNextValue(Utf8JsonReader& reader, ReadStack& state)
   --- End of inner exception stack trace ---
   at System.Text.Json.ThrowHelper.ReThrowWithPath(ReadStack& state, JsonReaderException ex)
   at System.Text.Json.JsonSerializer.GetReaderScopedToNextValue(Utf8JsonReader& reader, ReadStack& state)
   at System.Text.Json.JsonSerializer.Read[TValue](Utf8JsonReader& reader, JsonTypeInfo`1 jsonTypeInfo)
   at Anthropic.Models.Messages.RawMessageStreamEventConverter.Read(Utf8JsonReader& reader, Type typeToConvert, JsonSerializerOptions options) in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Models/Messages/RawMessageStreamEvent.cs:line 415
   at System.Text.Json.Serialization.JsonConverter`1.ReadCore(Utf8JsonReader& reader, T& value, JsonSerializerOptions options, ReadStack& state)
   at System.Text.Json.JsonSerializer.ReadFromSpan[TValue](ReadOnlySpan`1 json, JsonTypeInfo`1 jsonTypeInfo)
   at Anthropic.Core.Sse.Enumerate[T](HttpResponseMessage response, CancellationToken cancellationToken)+MoveNext()
   --- End of inner exception stack trace ---
   at Anthropic.Core.Sse.Enumerate[T](HttpResponseMessage response, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Core/Sse.cs:line 82
   at Anthropic.Core.Sse.Enumerate[T](HttpResponseMessage response, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Core/Sse.cs:line 26
   at Anthropic.Core.Sse.Enumerate[T](HttpResponseMessage response, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Anthropic.Services.MessageServiceWithRawResponse.<>c__DisplayClass7_0.<<CreateStreaming>g__Enumerate|2>d.MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Services/MessageService.cs:line 200
--- End of stack trace from previous location ---
   at Anthropic.Services.MessageServiceWithRawResponse.<>c__DisplayClass7_0.<<CreateStreaming>g__Enumerate|2>d.MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Services/MessageService.cs:line 200
--- End of stack trace from previous location ---
   at Anthropic.Services.MessageServiceWithRawResponse.<>c__DisplayClass7_0.<<CreateStreaming>g__Enumerate|2>d.System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult(Int16 token)
   at Anthropic.Core.StreamingHttpResponse`1.Enumerate(CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Core/HttpResponse.cs:line 188
   at Anthropic.Core.StreamingHttpResponse`1.Enumerate(CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Core/HttpResponse.cs:line 188
   at Anthropic.Core.StreamingHttpResponse`1.Enumerate(CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Anthropic.Services.MessageService.CreateStreaming(MessageCreateParams parameters, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Services/MessageService.cs:line 69
   at Anthropic.Services.MessageService.CreateStreaming(MessageCreateParams parameters, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/Services/MessageService.cs:line 69
   at Anthropic.Services.MessageService.CreateStreaming(MessageCreateParams parameters, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Microsoft.Extensions.AI.AnthropicClientExtensions.AnthropicChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/AnthropicClientExtensions.cs:line 598
   at Microsoft.Extensions.AI.AnthropicClientExtensions.AnthropicChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext() in /home/runner/work/anthropic-sdk-csharp/anthropic-sdk-csharp/src/Anthropic/AnthropicClientExtensions.cs:line 598
   at Microsoft.Extensions.AI.AnthropicClientExtensions.AnthropicChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Microsoft.Extensions.AI.FunctionInvokingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
   at Microsoft.Extensions.AI.FunctionInvokingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
   at Microsoft.Extensions.AI.FunctionInvokingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult()
   at Microsoft.Extensions.AI.LoggingChatClient.GetStreamingResponseAsync(IEnumerable`1 messages, ChatOptions options, CancellationToken cancellationToken)+MoveNext()
```

### DIFF
--- a/src/Anthropic.Tests/SseTest.cs
+++ b/src/Anthropic.Tests/SseTest.cs
@@ -223,7 +223,7 @@ public class SseTest : TestBase
     [Fact]
     public async Task Sse_SkipsDoneTerminalEvent()
     {
-        // Microsoft Foundry sends "data: [DONE]" (without an event: field) after
+        // Azure APIM may send "data: [DONE]" (without an event: field) after
         // message_stop to signal end-of-stream. SseParser assigns it the default
         // EventType "message", which matches the deserialization case. Without the
         // [DONE] guard this would throw a JsonException.
@@ -262,7 +262,7 @@ public class SseTest : TestBase
         }
 
         // Should have received 6 events (start, block_start, delta, block_stop, message_delta, message_stop)
-        // and NOT thrown on [DONE]
+        // and NOT throw on [DONE]
         Assert.Equal(6, actualMessages.Count);
     }
 }
--- a/src/Anthropic.Tests/SseTest.cs
+++ b/src/Anthropic.Tests/SseTest.cs
@@ -219,4 +219,50 @@ public class SseTest : TestBase
         Assert.IsType<AnthropicServiceException>(sseEx, exactMatch: false);
         Assert.Equal(ErrorType.OverloadedError, sseEx.ErrorType);
     }
+
+    [Fact]
+    public async Task Sse_SkipsDoneTerminalEvent()
+    {
+        // Microsoft Foundry sends "data: [DONE]" (without an event: field) after
+        // message_stop to signal end-of-stream. SseParser assigns it the default
+        // EventType "message", which matches the deserialization case. Without the
+        // [DONE] guard this would throw a JsonException.
+        var events = """
+            event: message_start
+            data: {"type":"message_start","message":{"id":"msg_01","type":"message","role":"assistant","model":"claude-haiku-4-5","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":0}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello"}}
+
+            event: content_block_stop
+            data: {"type":"content_block_stop","index":0}
+
+            event: message_delta
+            data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null},"usage":{"output_tokens":5}}
+
+            event: message_stop
+            data: {"type":"message_stop"}
+
+            data: [DONE]
+
+
+            """;
+
+        var resp = new HttpResponseMessage() { Content = new StringContent(events) };
+
+        var actualMessages = new List<JsonElement>();
+        await foreach (
+            var message in Sse.Enumerate<JsonElement>(resp, TestContext.Current.CancellationToken)
+        )
+        {
+            actualMessages.Add(message);
+        }
+
+        // Should have received 6 events (start, block_start, delta, block_stop, message_delta, message_stop)
+        // and NOT thrown on [DONE]
+        Assert.Equal(6, actualMessages.Count);
+    }
 }
--- a/src/Anthropic/Core/Sse.cs
+++ b/src/Anthropic/Core/Sse.cs
@@ -72,6 +72,13 @@ static class Sse
                 case "session.thread_status_rescheduled":
                 case "session.thread_status_terminated":
                 case "system.message":
+                    if (item.Data == "[DONE]")
+                    {
+                        // This is a terminal event sent by Microsoft Foundry to indicate the
+                        // end of the stream. It should not be treated as a message and should
+                        // not be deserialized.
+                        continue;
+                    }
                     T? message;
                     try
                     {
--- a/src/Anthropic/Core/Sse.cs
+++ b/src/Anthropic/Core/Sse.cs
@@ -74,9 +74,9 @@ static class Sse
                 case "system.message":
                     if (item.Data == "[DONE]")
                     {
-                        // This is a terminal event sent by Microsoft Foundry to indicate the
-                        // end of the stream. It should not be treated as a message and should
-                        // not be deserialized.
+                        // This is a terminal event that may be sent by Azure APIM to 
+                        // indicate the end of the stream. It should not be treated as
+                        // a message and should not be deserialized.
                         continue;
                     }
                     T? message;


dtmeadows-ant:
Hey @brused27 thanks for the proposal! I tried to reproduce this and so far haven't been able to trigger the `[DONE]` frame, so I'd love a bit more detail about your setup.

What I tested:
- Anthropic API (api.anthropic.com/v1/messages, stream: true): every SSE frame has a named event: field and the stream ends with event: message_stop — no `data: [DONE]` sentinel.
- Azure AI Foundry (https://<resource>.services.ai.azure.com/anthropic/v1/messages, East US 2, claude-opus-4-7): same shape — named events, terminates at message_stop, no [DONE]. Also tried several api-version query params. Streaming via AnthropicFoundryClient.Messages.CreateStreaming (without this PR's guard) completes cleanly.

The `[DONE]` sentinel isn't one of our conventions that I'm aware of.

Could you share:
1. Which endpoint/base URL the SDK was pointed at (e.g. `AnthropicFoundryClient` with a resource name, or a custom base URL)?
2. The deployment type and region for the Anthropic model (Global Standard / Data Zone / serverless, Model Router?), and the exact model/deployment name?
3. Whether there's anything in front of the endpoint (APIM policy, gateway, proxy) that might re-emit the stream?
4. If possible, the raw SSE tail showing the `[DONE]` frame, e.g. `curl -N https://<endpoint>/anthropic/v1/messages ... | tail -5` — mainly to confirm whether it arrives with or without an event: field.

brused27:
Hey @dtmeadows-ant - thank you for the review and you are spot on here that it does seem to be due to us passing through an Azure APIM now that I troubleshot this further.  We currently use this enterprise wide to manage our endpoints with subscription keys to make sure our LLM traffic is attributed properly against business units and ensure high availability cross-region for some scenarios we need if we exceed TPM limits.  However, we have configured this to be a passthrough policy ultimately so that existing Claude based apps will work properly (Claude Code, Anthropic SDK, etc).

> 1. Which endpoint/base URL the SDK was pointed at (e.g. AnthropicFoundryClient with a resource name, or a custom base URL)?

We are using `AnthropicClient` with BaseUrl and ApiKey supplied -  I redacted the hostnames, but we are calling against Azure APIM at: {REMOVED}.azure-api.net

```csharp
new AnthropicClient(new() { BaseUrl = endpoint, ApiKey = apiKey });
```

> 2. The deployment type and region for the Anthropic model (Global Standard / Data Zone / serverless, Model Router?), and the exact model/deployment name?

Global Standard, `claude-haiku-4-5` deployed with the same name as `claude-haiku-4-5` in Foundry is what is seen in the tests below.  This is a problem on any deployed Anthropic model for us, though.

> 3. Whether there's anything in front of the endpoint (APIM policy, gateway, proxy) that might re-emit the stream?

Yes, Azure APIM policy.

> 4. If possible, the raw SSE tail showing the [DONE] frame, e.g. curl -N https://<endpoint>/anthropic/v1/messages ... | tail -5 — mainly to confirm whether it arrives with or without an event: field.

Included below - Pure Microsoft Foundry vs Azure APIM in my environment curl request/responses.  I also included our Azure APIM policy below, which is purely a passthrough policy as seen below.


I did find that the way the SSE processing code works in this SDK currently, it drops back to take a default SSE structured event as assuming `event: message` per the .NET SDK so `item.EventType` is set to `message` even though no `event:` is seen on the response.  With `item.EventType` set to `message`, it ultimately causes it to fall through the switch statement with the `item.Data` parsed as `[DONE]` inside the SseParser enumerator; this is clearly not JSON compliant and ultimately what prompted the suggested PR fix here.  I realize now my comments/PR Title/Etc. are technically inaccurate now (referring to this as Microsoft Foundry, but technically it is Azure APIM upon further discovery). 😅

My hope is that we can bring it back in because this SDK works properly on v12.11.0 but the SSE parsing code seems to have been rewritten more recently, and it has broken our integration when using the pure SDK now.

### Azure APIM (stream DOES contain `data: [DONE]`)
```
curl --request POST \
  --url https://{REMOVED}.azure-api.net/anthropic/v1/messages \
  --header 'anthropic-version: 2023-06-01' \
  --header 'content-type: application/json' \
  --header 'x-api-key: {REMOVED}' \
  --data '{
  "model": "claude-haiku-4-5",
  "max_tokens": 1024,
  "stream": true,
  "system": "You are a helpful assistant.",
  "messages": [
    {
      "role": "user",
      "content": "Hello"
    }
  ]
}'
event: message_start
data: {"type":"message_start","message":{"model":"claude-haiku-4-5-20251001","id":"msg_{REMOVED}","type":"message","role":"assistant","content":[],"stop_reason":null,"stop_sequence":null,"stop_details":null,"usage":{"input_tokens":14,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":3,"service_tier":"standard","inference_geo":"not_available"}}           }

event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}      }

event: ping
data: {"type": "ping"}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello! "}    }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"👋 How can I help you today?"}          }

event: content_block_stop
data: {"type":"content_block_stop","index":0  }

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null,"stop_details":null},"usage":{"input_tokens":14,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":16}           }

event: message_stop
data: {"type":"message_stop"    }

data: [DONE]

```

### Microsoft Foundry (stream does NOT contain `data: [DONE]`)
```
curl --request POST \
  --url https://{REMOVED}.services.ai.azure.com/anthropic/v1/messages \
  --header 'anthropic-version: 2023-06-01' \
  --header 'content-type: application/json' \
  --header 'x-api-key: {REMOVED}' \
  --data '{
  "model": "claude-haiku-4-5",
  "max_tokens": 1024,
  "stream": true,
  "system": "You are a helpful assistant.",
  "messages": [
    {
      "role": "user",
      "content": "Hello"
    }
  ]
}'
event: message_start
data: {"type":"message_start","message":{"model":"claude-haiku-4-5-20251001","id":"msg_{REMOVED}","type":"message","role":"assistant","content":[],"stop_reason":null,"stop_sequence":null,"stop_details":null,"usage":{"input_tokens":14,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"cache_creation":{"ephemeral_5m_input_tokens":0,"ephemeral_1h_input_tokens":0},"output_tokens":3,"service_tier":"standard","inference_geo":"not_available"}}         }

event: content_block_start
data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}       }

event: ping
data: {"type": "ping"}

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello! "}           }

event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"👋 How can I help you today?"}     }

event: content_block_stop
data: {"type":"content_block_stop","index":0}

event: message_delta
data: {"type":"message_delta","delta":{"stop_reason":"end_turn","stop_sequence":null,"stop_details":null},"usage":{"input_tokens":14,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":16} }

event: message_stop
data: {"type":"message_stop"        }

```

### Azure APIM policy (passthrough)
```xml
<policies>
    <inbound>
        <base />
        <set-backend-service id="apim-generated-policy" backend-id="{REMOVED}" />
        <authentication-managed-identity resource="https://cognitiveservices.azure.com" />
    </inbound>
    <backend>
        <base />
    </backend>
    <outbound>
        <base />
    </outbound>
    <on-error>
        <base />
    </on-error>
</policies>
```

brused27:
As a follow up comparison, I am finding the current Anthropic Python SDK is gracefully handling this scenario against the APIM - I am hoping the .NET SDK can be re-aligned to that behavior the same way.

Here is a small sample script I am finding is succeeding against Azure APIM without error:
```python
import os
from anthropic import Anthropic

client = Anthropic(api_key=os.getenv("APIM_API_KEY"), base_url="https://{REMOVED}.azure-api.net/anthropic")
    
print("Response from Claude:")
    
with client.messages.stream(
    model="claude-haiku-4-5",
    max_tokens=1024,
    messages=[
        {"role": "user", "content": "hello"}
    ]
) as stream:
    for text in stream.text_stream:
        print(text, end="", flush=True)
    
print()
```

cc @dtmeadows-ant

feiiiiii5 (commented):
This thread has an unresolved question — the `[DONE]` sentinel "isn't one of our conventions", and the answer offered was that the Python SDK tolerates the same APIM. I checked that against the Python SDK, and it holds, but not for the reason assumed here. That changes what "align with Python" would mean, so it may be useful for the decision.

## The Python SDK has no `[DONE]` handling at all

`anthropic` 1.7.0: **zero** occurrences of the literal in the installed package. What it has is a decoder that leaves the event name unset plus a dispatcher that is a positive allowlist:

- `SSEDecoder.__init__` sets `self._event = None`, and `decode()` builds `ServerSentEvent(event=self._event, ...)` (`_streaming.py:377`, `:435-441`). Python does **not** apply the WHATWG default event type, so a frame with no `event:` field has `event=None`.
- `Stream.__stream__` yields only when `sse.event` equals one of the known names (`_streaming.py:71-126`). `None` matches nothing, so the frame is dropped without being parsed.

Running the real decoder over this PR's own fixture bytes, plus a probe frame that is unnamed but *not* `[DONE]`:

```
event='message_start'  data='{"type":"message_start",...}'                   -> deserialized
event='message_stop'   data='{"type":"message_stop"}'                         -> deserialized
event=None             data='[DONE]'                                          -> silently ignored
event=None             data='{"note":"an unnamed frame that is NOT [DONE]"}'   -> silently ignored
event='ping'           data='{}'                                              -> ignored (explicit `continue`)
```

Line four is the point: Python tolerates **any unnamed frame**, not this sentinel. Its behaviour falls out of two independent choices rather than a policy about stream terminators.

## Why .NET differs

`System.Net.ServerSentEvents.SseParser` does apply the spec default, so an unnamed frame arrives as `EventType == "message"` — and `"message"` is one of the labels in this switch (`Sse.cs:37`). The frame therefore matches the allowlist, reaches `JsonSerializer.Deserialize<T>` at `Sse.cs:80`, and produces exactly what the report shows. I reproduced it on `main` at `c4a1b81e` with only this PR's test added:

```
Anthropic.Tests.SseTest.Sse_SkipsDoneTerminalEvent [FAIL]
---- System.Text.Json.JsonException : 'D' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
   at Anthropic.Core.Sse.Enumerate[T](...) in src/Anthropic/Core/Sse.cs:line 87
   at Anthropic.Core.Sse.Enumerate[T](...) in src/Anthropic/Core/Sse.cs:line 80
Failed: 1, Passed: 0, Total: 1
```

At the PR head `9a00df92` the same test passes (1 passed, 0 failed). So the test bites and `main` has no `[DONE]` handling — this is not superseded.

The two SDKs disagree about the **default event type**, and everything else follows from that.

## Which means the guard is the pragmatic option, and it is correctly placed

`SseParserItem` exposes `EventType`, `Id`, `Data` and `Retry` — there is no way to ask "was an `event:` field present". So Python's rule ("ignore unnamed frames") is not expressible here, and a special case is what's left. I checked the placement specifically, because a guard written after a `case` group that did not include `"message"` would be dead code: `"message"` is at `Sse.cs:37`, inside the same group the guard lands in, and the stack trace above confirms the frame really does reach that block.

**One thing worth putting in the comment, though.** `item.Data == "[DONE]"` covers this sentinel only. Any other unnamed frame from a proxy — a `data: {"ping":1}` keep-alive, a different terminator spelling, injected telemetry — still reaches the deserializer and still throws, while Python ignores it. Not an argument against merging; just a reason for the comment to say "Azure APIM's `[DONE]` terminator" rather than implying unnamed frames are handled generally. If the maintainers would rather match Python exactly, the only lever I can see is tolerating a deserialization failure for `"message"` frames as a class, which is a wider behaviour change than this bug and would also swallow genuinely malformed `event: message` frames.

## Two mechanical items

**Base branch.** This targets `next` (`05fd63cf`). CLAUDE.md:10 is explicit about that case: *"PRs target the default branch (`main` on the public repo). `next` belongs to release automation — don't open PRs against it or push to it; a commit that lands there outside the release flow breaks the sync. Rebase onto the current default branch before asking for review."* `main` needs the change too — it has no `[DONE]` handling. The `Sse.cs` hunk retargets cleanly: its context lines are `Sse.cs:72-76` on current `main`, unchanged. The `SseTest.cs` hunk does **not** — `git apply` fails at `SseTest.cs:219` because that file has diverged on `main` since this branch was cut, so the test has to be re-added rather than cherry-picked. I hit that while reproducing and ended up inserting the method by hand.

**Nits, both optional.** `item.Data == "[DONE]"` is an exact match, so a trailing space — easy for a proxy to introduce — would miss; `Trim()` costs nothing. And the guard sits inside a 46-label `case` group, so it adds a string comparison to every event; hoisting it above the `switch` would make it obvious that it applies to all of them.

Neither of these is a reason to hold the PR. The underlying report is real, the fix is in the right place, and the test demonstrably fails without it.

Reproduce: worktree at each sha, `dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj -f net8.0 --filter "FullyQualifiedName~Sse_SkipsDoneTerminalEvent"`, .NET SDK 9.0.318, macOS arm64. Python side: `anthropic==1.7.0`, `SSEDecoder().iter_bytes([...])` over the fixture bytes.

AI-assisted review; the measurements above are mine.

brused27:
> This thread has an unresolved question — the `[DONE]` sentinel "isn't one of our conventions", and the answer offered was that the Python SDK tolerates the same APIM. I checked that against the Python SDK, and it holds, but not for the reason assumed here. That changes what "align with Python" would mean, so it may be useful for the decision.
> 
> ## The Python SDK has no `[DONE]` handling at all
> `anthropic` 1.7.0: **zero** occurrences of the literal in the installed package. What it has is a decoder that leaves the event name unset plus a dispatcher that is a positive allowlist:
> 
> * `SSEDecoder.__init__` sets `self._event = None`, and `decode()` builds `ServerSentEvent(event=self._event, ...)` (`_streaming.py:377`, `:435-441`). Python does **not** apply the WHATWG default event type, so a frame with no `event:` field has `event=None`.
> * `Stream.__stream__` yields only when `sse.event` equals one of the known names (`_streaming.py:71-126`). `None` matches nothing, so the frame is dropped without being parsed.
> 
> Running the real decoder over this PR's own fixture bytes, plus a probe frame that is unnamed but _not_ `[DONE]`:
> 
> ```
> event='message_start'  data='{"type":"message_start",...}'                   -> deserialized
> event='message_stop'   data='{"type":"message_stop"}'                         -> deserialized
> event=None             data='[DONE]'                                          -> silently ignored
> event=None             data='{"note":"an unnamed frame that is NOT [DONE]"}'   -> silently ignored
> event='ping'           data='{}'                                              -> ignored (explicit `continue`)
> ```
> 
> Line four is the point: Python tolerates **any unnamed frame**, not this sentinel. Its behaviour falls out of two independent choices rather than a policy about stream terminators.
> 
> ## Why .NET differs
> `System.Net.ServerSentEvents.SseParser` does apply the spec default, so an unnamed frame arrives as `EventType == "message"` — and `"message"` is one of the labels in this switch (`Sse.cs:37`). The frame therefore matches the allowlist, reaches `JsonSerializer.Deserialize<T>` at `Sse.cs:80`, and produces exactly what the report shows. I reproduced it on `main` at `c4a1b81e` with only this PR's test added:
> 
> ```
> Anthropic.Tests.SseTest.Sse_SkipsDoneTerminalEvent [FAIL]
> ---- System.Text.Json.JsonException : 'D' is an invalid start of a value. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
>    at Anthropic.Core.Sse.Enumerate[T](...) in src/Anthropic/Core/Sse.cs:line 87
>    at Anthropic.Core.Sse.Enumerate[T](...) in src/Anthropic/Core/Sse.cs:line 80
> Failed: 1, Passed: 0, Total: 1
> ```
> 
> At the PR head `9a00df92` the same test passes (1 passed, 0 failed). So the test bites and `main` has no `[DONE]` handling — this is not superseded.
> 
> The two SDKs disagree about the **default event type**, and everything else follows from that.
> 
> ## Which means the guard is the pragmatic option, and it is correctly placed
> `SseParserItem` exposes `EventType`, `Id`, `Data` and `Retry` — there is no way to ask "was an `event:` field present". So Python's rule ("ignore unnamed frames") is not expressible here, and a special case is what's left. I checked the placement specifically, because a guard written after a `case` group that did not include `"message"` would be dead code: `"message"` is at `Sse.cs:37`, inside the same group the guard lands in, and the stack trace above confirms the frame really does reach that block.
> 
> **One thing worth putting in the comment, though.** `item.Data == "[DONE]"` covers this sentinel only. Any other unnamed frame from a proxy — a `data: {"ping":1}` keep-alive, a different terminator spelling, injected telemetry — still reaches the deserializer and still throws, while Python ignores it. Not an argument against merging; just a reason for the comment to say "Azure APIM's `[DONE]` terminator" rather than implying unnamed frames are handled generally. If the maintainers would rather match Python exactly, the only lever I can see is tolerating a deserialization failure for `"message"` frames as a class, which is a wider behaviour change than this bug and would also swallow genuinely malformed `event: message` frames.
> 
> ## Two mechanical items
> **Base branch.** This targets `next` (`05fd63cf`). CLAUDE.md:10 is explicit about that case: _"PRs target the default branch (`main` on the public repo). `next` belongs to release automation — don't open PRs against it or push to it; a commit that lands there outside the release flow breaks the sync. Rebase onto the current default branch before asking for review."_ `main` needs the change too — it has no `[DONE]` handling. The `Sse.cs` hunk retargets cleanly: its context lines are `Sse.cs:72-76` on current `main`, unchanged. The `SseTest.cs` hunk does **not** — `git apply` fails at `SseTest.cs:219` because that file has diverged on `main` since this branch was cut, so the test has to be re-added rather than cherry-picked. I hit that while reproducing and ended up inserting the method by hand.
> 
> **Nits, both optional.** `item.Data == "[DONE]"` is an exact match, so a trailing space — easy for a proxy to introduce — would miss; `Trim()` costs nothing. And the guard sits inside a 46-label `case` group, so it adds a string comparison to every event; hoisting it above the `switch` would make it obvious that it applies to all of them.
> 
> Neither of these is a reason to hold the PR. The underlying report is real, the fix is in the right place, and the test demonstrably fails without it.
> 
> Reproduce: worktree at each sha, `dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj -f net8.0 --filter "FullyQualifiedName~Sse_SkipsDoneTerminalEvent"`, .NET SDK 9.0.318, macOS arm64. Python side: `anthropic==1.7.0`, `SSEDecoder().iter_bytes([...])` over the fixture bytes.
> 
> AI-assisted review; the measurements above are mine.

Wow, thank you so much for the in-depth review.  You are refreshing my memory, and you definitely picked up on the exact difference here that is causing this to be problematic in .NET but not in Python.  SSE processing by .NET follows the SSE/RFC spec tightly where if a data row is sent but no event row was included, it will default to being treated as a `message` event.  However, Python's implementation seems to ignore this and clearly causes a different final outcome with this situation in these SDKs.  I primarily provided the python sample as an additional datapoint.  My goal with this small PR is to tighten handling to avoid unnecessary exception throws while also working toward avoiding other potential situations where we truly might see a deserialization error.

I have updated the merge destination to `main`; prior guidance at time of submission back in May was to request PR's against the next branch, but it looks like that guidance has changed.

Please note that I am working around this problem in our platform's code by wrapping the async return stream to catch the problematic event Exception and simulating a successful response. Unfortunately, this requires Exception catches with deeper inspection; I would prefer a more holistic and performant solution like what I proposed so that we don't require workarounds on the streaming side to stay current on the SDK. The workaround I have implemented is one of the reasons I have not pushed further for this fix, but if it is acceptable to the maintainers of the SDK, I think it would be helpful for any others who are hosting these models behind an Azure APIM.

PR #204, opened 2026-05-28
fix: emit only the completed tool in streaming content_block_stop handler

### LINKED ISSUE #197: GetStreamingResponseAsync emits duplicate/incomplete FunctionCallContent when multiple tools are called in parallel

Sarmkadan:
## Summary

Fixes #197 — `GetStreamingResponseAsync` emits duplicate/incomplete `FunctionCallContent` when multiple tools are called in parallel.

### Root cause

`RawContentBlockStopEvent` iterates **all** entries in `streamingFunctions` and clears the dictionary on every `content_block_stop` event. When tools are called in parallel:

1. The first `content_block_stop` emits every tracked tool — including those still receiving `input_json_delta` chunks — with potentially incomplete arguments
2. Subsequent `content_block_stop` events find an empty dictionary and emit nothing
3. `FunctionInvokingChatClient` constructs a response with duplicate `tool_use` ids, causing the API to reject with `"tool_use ids must be unique"`

### Fix

Replace the `foreach` + `Clear()` with a targeted `Dictionary.Remove()` at `contentBlockStop.Index`. Each tool is now emitted exactly once, only when its own stop event fires, with fully accumulated arguments.

The same fix is applied to the Beta streaming path in `AnthropicBetaClientExtensions.cs`.

### Test

Added `GetStreamingResponseAsync_WithParallelToolCalls_EmitsOnlyAtOwnStopEvent` — a regression test where one tool's `content_block_stop` fires while another tool still has pending `input_json_delta` chunks. This test fails on `main` and passes with the fix.

## Test plan

- [x] New regression test covers the exact interleaving scenario from the issue
- [x] All existing streaming tool call tests pass (sequential, interleaved, fragmented, parameterless, server tool use)
- [x] Full test suite: 9269 passed, 0 regressions (52 pre-existing failures are API connectivity tests)

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -3061,6 +3061,90 @@ public abstract class AnthropicClientExtensionsTestsBase
     }
 
     [Fact]
+    public async Task GetStreamingResponseAsync_WithParallelToolCalls_EmitsOnlyAtOwnStopEvent()
+    {
+        // Regression test: when tool A stops while tool B still has pending deltas,
+        // only tool A should be emitted — not both.
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Call parallel tools"
+                    }]
+                }],
+                "stream": true
+            }
+            """,
+            actualResponse: """
+            event: message_start
+            data: {"type":"message_start","message":{"id":"msg_par_01","type":"message","role":"assistant","model":"claude-haiku-4-5","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":10,"output_tokens":0}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":0,"content_block":{"type":"tool_use","id":"toolu_fast","name":"get_weather","input":{},"caller":{"type":"direct"}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":1,"content_block":{"type":"tool_use","id":"toolu_slow","name":"search_db","input":{},"caller":{"type":"direct"}}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":0,"delta":{"type":"input_json_delta","partial_json":"{\"city\":\"Berlin\"}"}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":1,"delta":{"type":"input_json_delta","partial_json":"{\"table\":"}}
+
+            event: content_block_stop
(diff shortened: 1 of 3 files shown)


TomerAberbach:
Seems like a duplicate of https://github.com/anthropics/anthropic-sdk-csharp/pull/200

PR #196, opened 2026-05-14
fix: preserve top-level json schema keywords

danielsogl:
Fixes #195.

## Summary

When converting `ChatResponseFormatJson` to `OutputConfig.Format.JsonOutputFormat.Schema`, the previous code constructed a new dictionary with only four keys (`type`, `properties`, `required`, `additionalProperties`), silently dropping every other top-level JSON Schema keyword — notably `$defs` and `definitions`, which Anthropic's structured-outputs API explicitly supports for internal `$ref` resolution ([docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)).

For schemas that use `$ref` to reference internal `$defs`, the request was rejected by the server with:

```
HTTP 400 — AnthropicBadRequestException
{"type":"error","error":{"type":"invalid_request_error",
 "message":"output_config.format.schema: Invalid schema:
            Reference to non-existent definition: #/$defs/..."}}
```

## Fix

Preserve all top-level keywords of the transformed schema and only override `type` and `additionalProperties` to satisfy API requirements:

```csharp
var schemaDict = new Dictionary<string, JsonElement>(StringComparer.Ordinal);
foreach (JsonProperty p in schema.EnumerateObject())
{
    schemaDict[p.Name] = p.Value;
}
schemaDict["type"] = JsonElement.Parse("\"object\"");
schemaDict["additionalProperties"] = JsonElement.Parse("false");
```

This is a strict super-set of the current behavior for schemas without `$defs`.

## Test

Added `GetResponseAsync_WithResponseFormatSchemaContainingDefs_PreservesDefs` which verifies that a schema with `$defs` round-trips through the adapter intact to `output_config.format.schema`.

All 180 existing `AnthropicClientExtensions` tests still pass.

PR #205, opened 2026-05-28
fix: preserve $defs and $ref in ChatResponseFormatJson schema mapping

### LINKED ISSUE #195: AnthropicChatClient drops $defs / $ref when forwarding ChatResponseFormatJson.Schema to OutputConfig.Format
## Description

When converting `ChatOptions.ResponseFormat` (a `ChatResponseFormatJson` with a `Schema`) into the Anthropic `OutputConfig.Format`, the `AnthropicChatClient` adapter constructs a new schema dictionary with only four top-level keys: `type`, `properties`, `required`, `additionalProperties`. Any other top-level JSON Schema keywords — most notably `$defs` — are silently stripped.

For schemas that use `$ref` to reference internal `$defs` (which Anthropic's structured outputs explicitly supports, per [the docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)), the request is then rejected by the server:

```
HTTP 400 — AnthropicBadRequestException
{"type":"error","error":{"type":"invalid_request_error",
 "message":"output_config.format.schema: Invalid schema:
            Reference to non-existent definition: #/$defs/Segments"}}
```

## Reproduction

```csharp
using Anthropic;
using Microsoft.Extensions.AI;
using System.Text.Json;

var schemaJson = """
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "value": { "$ref": "#/$defs/Item" }
  },
  "required": ["value"],
  "$defs": {
    "Item": { "type": "string" }
  }
}
""";

IChatClient chatClient = new AnthropicClient { ApiKey = "..." }
    .AsIChatClient("claude-sonnet-4-5-20250929");

var response = await chatClient.GetResponseAsync(
    [new ChatMessage(ChatRole.User, "Return any value")],
    new ChatOptions
    {
        ResponseFormat = ChatResponseFormat.ForJsonSchema(
            JsonSerializer.Deserialize<JsonElement>(schemaJson),
            schemaName: "Result"),
    });
// Throws AnthropicBadRequestException
```

## Root Cause

[`src/Anthropic/AnthropicClientExtensions.cs`](https://github.com/anthropics/anthropic-sdk-csharp/blob/main/src/Anthropic/AnthropicClientExtensions.cs) (around the `CreateMessageParams` schema-mapping branch):

```csharp
case ChatResponseFormatJson formatJson when formatJson.Schema is not null:
    JsonElement schema = JsonSchemaTransformCache
        .GetOrCreateTransformedSchema(formatJson).GetValueOrDefault();
    if (schema.TryGetProperty("properties", out var properties)
        && properties.ValueKind is JsonValueKind.Object
        && schema.TryGetProperty("required", out var required)
        && required.ValueKind is JsonValueKind.Array)
    {
        createParams = createParams with
        {
            OutputConfig = (createParams.OutputConfig ?? new OutputConfig()) with
            {
                Format = new JsonOutputFormat()
                {
                    Schema = new Dictionary<string, JsonElement>
                    {
                        ["type"] = JsonElement.Parse("\"object\""),
                        ["properties"] = properties,
                        ["required"] = required,
                        ["additionalProperties"] = JsonElement.Parse("false"),
                    },
                },
            },
        };
    }
```

The reconstructed dictionary contains only four keys. `$defs`, `definitions`, `$schema`, `title`, `description`, and any other valid top-level JSON Schema property are dropped.

## Expected Behavior

The adapter should forward all top-level JSON Schema properties that Anthropic's structured-outputs endpoint accepts. At minimum, `$defs` (and `definitions` for older drafts), since the documentation explicitly states they are supported:

> Supported features: `$ref`, `$def`, and `definitions` (external `$ref` not supported)
> — https://platform.claude.com/docs/en/build-with-claude/structured-outputs

## Proposed Fix

Instead of constructing a new dictionary with only four keys, clone the transformed `JsonElement`'s top-level object and only overwrite `additionalProperties`:

```csharp
var schemaDict = schema.EnumerateObject()
    .ToDictionary(p => p.Name, p => p.Value);
schemaDict["additionalProperties"] = JsonElement.Parse("false");

createParams = createParams with
{
    OutputConfig = (createParams.OutputConfig ?? new OutputConfig()) with
    {
        Format = new JsonOutputFormat { Schema = schemaDict },
    },
};
```

This preserves `$defs`, `description`, `title`, etc. and is a strict super-set of the current behavior for the common case (no `$defs`).

## Environment

| | |
|---|---|
| Anthropic SDK | 12.21.0 |
| Microsoft.Extensions.AI | 10.6.0 |
| .NET | 10.0 |
| Model | claude-sonnet-4-5-20250929, claude-sonnet-4-6 |

## Related Issues

- #109 (`output_format` → `output_config.format` migration) — closed
- #165 (`OutputConfig` merge with `RawRepresentationFactory`) — closed

Both addressed different aspects; the `$defs`-stripping is independent.

Sarmkadan:
## Summary

Fixes #195 — `AnthropicChatClient` drops `$defs` / `$ref` when forwarding `ChatResponseFormatJson.Schema` to `OutputConfig.Format`, causing `HTTP 400 — Reference to non-existent definition`.

### Root cause

The `OutputConfig.Format.Schema` dictionary was constructed with only four hardcoded keys:

```csharp
Schema = new Dictionary<string, JsonElement>
{
    ["type"] = ...,
    ["properties"] = properties,
    ["required"] = required,
    ["additionalProperties"] = ...,
}
```

Any other top-level JSON Schema keywords — most notably `$defs` — were silently stripped. Schemas that use `$ref` to reference internal `$defs` (which [Anthropic's structured outputs explicitly supports](https://platform.claude.com/docs/en/build-with-claude/structured-outputs)) were then rejected by the server.

### Fix

Iterate all properties from the already-transformed schema (which the `JsonSchemaTransformCache` has already filtered to only supported keywords) and copy them into the dictionary. Then `TryAdd` `type` and `additionalProperties` as defaults if not already present.

This preserves `$defs`, `$ref`, `anyOf`, `allOf`, and any future keywords the transform cache allows through. The same fix is applied to the Beta path in `AnthropicBetaClientExtensions.cs`.

### Test

Added `GetResponseAsync_WithDefsRefSchema_PreservesDefinitions` — verifies that a schema using `$defs`/`$ref` round-trips correctly through the `VerbatimHttpHandler` (exact JSON match against expected API request).

## Test plan

- [x] New test verifies `$defs`/`$ref` schema is forwarded to the API
- [x] All existing schema tests pass (simple, nested, enum, all-transformations)
- [x] Full test suite: 9269 passed, 0 regressions (52 pre-existing failures are API connectivity tests)

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -6161,6 +6161,112 @@ public abstract class AnthropicClientExtensionsTestsBase
     }
 
     [Fact]
+    public async Task GetResponseAsync_WithDefsRefSchema_PreservesDefinitions()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-sonnet-4-5-20250929",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Return a result"
+                    }]
+                }],
+                "output_config": {
+                    "format": {
+                        "type": "json_schema",
+                        "schema": {
+                            "type": "object",
+                            "properties": {
+                                "segments": {
+                                    "type": "array",
+                                    "items": { "$ref": "#/$defs/Segment" }
+                                }
+                            },
+                            "required": ["segments"],
+                            "$defs": {
+                                "Segment": {
+                                    "type": "object",
+                                    "properties": {
+                                        "label": { "type": "string" },
+                                        "value": { "type": "integer" }
+                                    },
+                                    "required": ["label", "value"]
+                                }
+                            },
+                            "additionalProperties": false
+                        }
+                    }
+                }
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_defs_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-sonnet-4-5-20250929",
+                "content": [{
+                    "type": "text",
+                    "text": "{\"segments\":[{\"label\":\"A\",\"value\":1}]}"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 20,
+                    "output_tokens": 10
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-sonnet-4-5-20250929");
+
+        ChatOptions options = new()
+        {
+            ResponseFormat = ChatResponseFormat.ForJsonSchema(
+                JsonElement.Parse(
+                    """
+                    {
+                        "type": "object",
+                        "properties": {
+                            "segments": {
+                                "type": "array",
+                                "items": { "$ref": "#/$defs/Segment" }
+                            }
+                        },
+                        "required": ["segments"],
+                        "$defs": {
+                            "Segment": {
+                                "type": "object",
+                                "properties": {
+                                    "label": { "type": "string" },
+                                    "value": { "type": "integer" }
+                                },
+                                "required": ["label", "value"]
+                            }
+                        }
+                    }
+                    """
+                ),
+                "SegmentResult"
+            ),
+        };
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Return a result",
+            options,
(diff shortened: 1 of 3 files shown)


TomerAberbach:
This was done in https://github.com/anthropics/anthropic-sdk-csharp/pull/196

PR #207, opened 2026-05-30
feat(meai): map mid-conversation system messages in place

PederHP:
### Summary

The MEAI `IChatClient` mapping hoisted every `ChatRole.System` message into the top-level `system` property regardless of position, so mid-conversation system messages were unreachable.

Now, leading system messages still map to the top-level `system` property (the only valid representation — the API rejects a leading `{"role":"system"}` even on supporting models). Every later `ChatRole.System` message is emitted as a `{"role":"system"}` message at its position, as-is. The SDK does not validate the placement or the model; the API rejects the request when either doesn't support mid-conversation system messages. Consecutive system messages are merged into one: the API treats adjacent system messages as a single system section, but some providers (e.g. Amazon Bedrock) reject them.

```csharp
List<ChatMessage> messages =
[
    new(ChatRole.System, "You are a helpful assistant."),   // top-level `system`
    new(ChatRole.User, "Draft the release notes."),
    new(ChatRole.System, "Keep the response under 100 words."),  // emitted in place
];

```

Applied to both the stable and beta `IChatClient` mappings — the two files' hunks are byte-identical modulo the `Beta` type prefix.

### Breaking change, per @sophie-ant's review

The previous revision was opt-in (`ChatOptions.WithInterleavedSystemMessages()`) and validated placements, silently hoisting non-qualifying system messages into the top-level `system` property. Per the review ("send it as-is and let it blow up rather than semantically changing the request"), the validation pass and the opt-in are both gone — the in-place mapping is the default and only behavior.

**BREAKING**: conversations containing mid-conversation `ChatRole.System` messages that previously relied on hoisting now fail on models without mid-conversation system message support.

Also addressed from the review:

- The incorrect claim that the API rejects consecutive system messages (it accepts them as a single section; Bedrock rejects them) — comment rewritten.
- The incorrect claims that the feature is unavailable on Bedrock/Vertex/Foundry and restricted to "Claude Opus 4.8 and later" — removed with the deleted extension file; no provider or model-version claims remain.
- `IsRole`'s string comparison — deleted along with the validation pass it served.

### Test plan

- `dotnet test` — 188 MEAI extension tests pass across the stable and beta clients (net8.0), including in-place emission after a user turn, between user and assistant turns, pass-through of an (invalid) system message after an assistant turn, consecutive-message merging, leading-message hoisting, and cache control on an in-place system message.
- `dotnet build` — clean on all target frameworks.
- `./scripts/lint` — csharpier check, `dotnet format style`, `dotnet format analyzers` all clean.

Co-Authored-By: Claude Opus 5 and Claude Fable 5 <noreply@anthropic.com>


src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1429,6 +1543,69 @@ existingSystem.Value is IReadOnlyList<BetaTextBlockParam> existingMessages
...
         }
 
+        /// <summary>
+        /// Determines whether the resolved model supports mid-conversation system messages.
+        /// This is a capability allowlist; extend it as additional models gain support. Note
+        /// that the feature is unavailable on Bedrock, Vertex, and Foundry regardless of model.
+        /// </summary>
+        private static bool ModelSupportsMidConversationSystem(MessageCreateParams createParams) =>
+            createParams.Model.Raw() is { } modelId
+            && modelId.StartsWith("claude-opus-4-8", StringComparison.Ordinal);
```
jozkee: Bedrock sends `anthropic.claude-opus-4-8`, so the prefix check misses it. We could trim `anthropic.` to fix this. Foundry, though, breaks even that: per the [docs](https://platform.claude.com/docs/en/build-with-claude/claude-in-microsoft-foundry#api-model-ids-and-deployments), model is the deployment name, which can be anything.

I haven't seen model-driven behavior in other IChatClients e.g. Microsoft.Extensions.AI.OpenAI or Google.GenAI.  I think is fine, but I suspect it would be a first.

PederHP: An alternative to the model-driven behavior would be a builder style WithInterleavedSystemMessages() on IChatClient or something passed in ChatOptions. Downside is developers having to know to use them. The ergonomics of having to check the model are a bit rough if the developer has to do it.

I can change to either of these if you prefer - or maybe add one and keep the current behavior , with the trim added. That provides a path for foundry users while keeping ergonomics convenient for those with the right model name?

jozkee: I like the idea of having an extension, I think a `WithInterleavedSystemMessages(this ChatOptions, InterleavedSystemMessageMode mode = Auto)` that adds an additional property, similar to WithCacheControl, should work.

Auto or unspecified can keep parsing the model id, while Enabled/Disabled provide an avenue for providers with custom model ids.

The extension is just for discoverability that a bare additional property would otherwise lack. Models that don't support this today will eventually phase out, at which point the extension becomes a no-op that we can then obsolete -> remove.

jozkee: Scratch Auto mode, it would be better to make this explicit only and avoid setting the model-driven precedent.

sophie-ant:
Is this PR able to be broken up at all? There's a lot in here and it makes it tricky to review. This'll need a rebase as well

PederHP:
I will do a rebase and look at simplifying it. It looks worse than it is though because of the missing rebase.

sophie-ant:
@PederHP FYI I am off next week but I'll take a look the following week!

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: Done — the model-id sniffing is gone entirely. `WithInterleavedSystemMessages(this ChatOptions options, bool enabled = true)` is now the only gate, so Bedrock's `anthropic.` prefix and arbitrary Foundry deployment names all work without the SDK trying to parse anything. There's a test using a `my-foundry-deployment` model id to pin that down.

No `Auto`, per your follow-up. With `Auto` off the table an `InterleavedSystemMessageMode` enum degenerates to two states, so I went with a plain `bool` — happy to switch to the enum if you'd rather keep room to re-add modes later without an overload.

Storage follows `WithCacheControl`: an `anthropic:`-prefixed key in `AdditionalProperties` with an `internal` getter. And as you noted, once the models that lack support phase out this becomes a no-op we can obsolete and remove.

Callers stay responsible for using a supporting model — the SDK doesn't verify it and the API errors if unsupported. That's documented on the method.

(above was authored by Claude)

PederHP:
Rebased onto `next`. The bulk of what made this hard to review was the missing rebase — the branch was sitting on four commits that had already landed upstream under different SHAs, so the diff showed ~100 files. It's now a single commit over four files: the two `IChatClient` mappings (byte-identical hunks modulo the `Beta` prefix), one new ~90-line extension file, and the shared test base.

I also addressed @jozkee's review in the same pass, which removed rather than added code: the model-id capability check is gone, replaced by an explicit `ChatOptions.WithInterleavedSystemMessages()` opt-in. That let me cut the long comment block down as well.

Enjoy your week off — no rush. Given how often `next` gets rewritten I expect this will need another rebase before it merges; happy to do that whenever you're ready to look.

(above authored with Claude)

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -549,6 +550,43 @@ out StreamingFunctionData? completedFunction
             }
         }
 
+        /// <summary>
+        /// Merges runs of consecutive <see cref="ChatRole.System"/> messages into a single
+        /// message. The API does not allow consecutive mid-conversation system messages, so
```
sophie-ant: FYI this is untrue in the general case: https://platform.claude.com/docs/en/build-with-claude/mid-conversation-system-messages

> Consecutive system messages are accepted and treated as a single system section, which follows the same placement rule as a whole.

It looks like Bedrock doesn't allow them so it's okay if the logic stays, though the comment should be updated

src/Anthropic/ChatOptionsSystemMessageExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+    /// <see langword="true"/> to emit qualifying mid-conversation system messages in place;
+    /// <see langword="false"/> to hoist every system message into the top-level <c>system</c>
+    /// property. The default is <see langword="false"/> when this method is not called.
+    /// </param>
+    /// <returns>The same <paramref name="options"/> instance for method chaining.</returns>
+    /// <remarks>
+    /// <para>
+    /// This is opt-in only: the SDK never infers it from the model id. Mid-conversation system
+    /// messages require a model that supports them (Claude Opus 4.8 and later), and are not
+    /// available on Bedrock, Vertex, or Foundry. The SDK does not verify either condition — the
```
sophie-ant: I don't think this is correct - it should work on Bedrock and Vertex at least

sophie-ant: I also wouldn't say `Claude Opus 4.8 and later` since this may not always hold

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1451,6 +1563,60 @@ existingSystem.Value is IReadOnlyList<BetaTextBlockParam> existingMessages
             return AddMeaiHeaders(createParams);
         }
 
+        private static bool IsRole(BetaMessageParam message, string role) =>
+            string.Equals(message.Role.Raw(), role, StringComparison.Ordinal);
```
sophie-ant: I think we should be doing an enum comparison with `message.Role.Value()` instead of a string comparison

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1420,6 +1485,53 @@ enabled with
...
+                bool interleavedSystemMessages = options.GetInterleavedSystemMessages();
+                IReadOnlyList<BetaMessageParam> builtMessages = createParams.Messages;
+                List<BetaMessageParam> rewritten = [];
+                bool hoistedAny = false;
+                for (int i = 0; i < builtMessages.Count; i++)
+                {
+                    BetaMessageParam message = builtMessages[i];
+                    if (TryGetSystemMessageBlocks(message, out var blocks))
+                    {
+                        bool validPlacement =
```
sophie-ant: Imo we are doing a little too much in here to try and figure out what the user wants, when we could just be passing it to the API and letting it blow up (so that the user has something to correct). I think silently changing the semantics of their request is confusing

sophie-ant (changes requested):
I think in general I am wondering if we can get away with doing less when trying to massage the request so it won't get rejected; where possible, we should send it "as-is" and let it blow up rather than semantically changing the request

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: Fixed — the comment now says the API accepts consecutive system messages and treats them as a single system section, and that the merge is kept because some providers (e.g. Amazon Bedrock) reject adjacent system messages. The merge logic itself is unchanged.

(above authored by Claude)

src/Anthropic/ChatOptionsSystemMessageExtensions.cs
PederHP: That claim is gone — `ChatOptionsSystemMessageExtensions.cs` was deleted entirely (see the thread on the validation block), and no provider-availability claims were reintroduced elsewhere.

(above authored by Claude)

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: `IsRole` (and `TryGetSystemMessageBlocks`, its only caller's helper) were deleted along with the whole post-hoc validation pass, so there's no role comparison left to convert. If a role check comes back in a future revision it'll use `message.Role.Value()` against the enum.

(above authored by Claude)

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: Agreed, and this reshaped the PR: the entire placement-validation/hoist pass in `GetMessageCreateParams` is gone, and with the behavior no longer conditional the `WithInterleavedSystemMessages` opt-in lost its purpose, so it's deleted too. The mapping is now: leading system messages go to the top-level `system` property (the only valid representation — the API rejects a leading `{"role":"system"}` even on supporting models); every later `ChatRole.System` message is emitted as a `{"role":"system"}` message at its position, as-is, and the API rejects the request if the model or the placement doesn't support it. The SDK no longer rewrites the request's semantics.

Note this makes the change breaking: conversations with mid-conversation system messages that previously relied on hoisting now fail on models without support. The commit is marked `feat(meai)!` with a `BREAKING CHANGE` footer. Net effect on the diff: −444 lines versus the previous revision.

(above authored by Claude)

PederHP:
@sophie-ant

Review comments addressed. I like the simpler direction you requested, but this does make it a breaking change, as mentioned above - which is why I went for the more complex solution at first.

The main cases where this will break existing code when used with a model that doesn't support interleaved system messages:

- System message placed at the end of the message array instead of as the first message.

- Cross-provider code where system message are interleaved to take advantage of providers that support interleaved messages.

There are probably some of those out there. I honestly think it is good to be explicit about re-ordering when the model doesn't support interleaving. With the current SDK, the messages are lifted into the property, and this means that semantics are silently being changed as you point out. Most developers don't know this and might expect the model to see the interleaving when it actually doesn't. So while it is a breaking change for some applications, I would say it is a good breaking change as it doesn't silently reorder messages behind the developers back, but simply fails if the order is invalid.

sophie-ant (approved):
Thanks so much! Sorry about the turnaround time!

PR #217, opened 2026-06-18
fix(helpers): rewrap delta.Usage.Iterations to BetaUsageIteration

dtmeadows-ant:
The C# generator (since commit 050cd3cc) inlines the `iterations` union under two distinct types — `Iteration` on `BetaMessageDeltaUsage` vs `BetaUsageIteration` on `BetaUsage` — and renamed the variant accessors to `TryPickBeta*`. Both reference the same schema; the shared `BetaIterationsUsageItems` type was deleted by the generator (same spec+config hash → generator behavior change).

- Rewrap `delta.Usage.Iterations` via raw `JsonElement` in the aggregator
- Update the streaming test to the new `TryPickBeta*` method names

Unblocks the release PR (#216). The shared-type restoration is a Stainless-side fix.

### DIFF
--- a/.release-please-manifest.json
+++ b/.release-please-manifest.json
@@ -1,5 +1,5 @@
 {
-  "src/Anthropic": "12.29.1",
+  "src/Anthropic": "12.30.0",
   "src/Anthropic.Foundry": "0.7.0",
   "src/Anthropic.Bedrock": "0.10.1",
   "src/Anthropic.Vertex": "0.5.0",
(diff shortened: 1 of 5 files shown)


dtmeadows-ant:
Closing — Tomer's taking the custom-code fix for the type split from stainless-api/stainless#18342 (https://ant-core-systems.slack.com/archives/C0B4NU9DKA8/p1781805635658329). The diff here is the same shape if useful as a reference: typed-switch rewrap of `delta.Usage.Iterations` → `BetaUsageIteration`, plus the `TryPickBeta*` rename in the streaming test.

PR #220, opened 2026-06-25
feat(meai): map HostedToolSearchTool to Anthropic tool search

jozkee:
Add support for Microsoft.Extensions.AI HostedToolSearchTool in the stable and beta IChatClient adapters. The tool defaults to the regex tool-search variant, with an AsBm25() extension to opt into BM25. 

Server tool-use and tool-search-result response blocks now round-trip back to their request-side *Param types.


sophie-ant:
Mind rebasing this?

jozkee:
@sophie-ant done.

sophie-ant:
Would you be able to fix the lint? FYI I am off next week but will take a look when I get back

sophie-ant:
Looks like there's still a lint failure

jozkee:
Fixed, it wasn't being flagged locally in Windows.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -864,6 +842,23 @@ out List<TextBlockParam>? systemMessages
                             contents.Add(rawContent);
                             break;
 
+                        case AIContent ac
+                            when ac.RawRepresentation is ServerToolUseBlock serverToolUseBlock:
+                            contents.Add(
+                                ServerToolUseBlockParam.FromRawUnchecked(serverToolUseBlock.RawData)
+                            );
+                            break;
```
tomer-ant, line 957: This case matches every ServerToolUseBlock, not only the tool-search ones, but the only result block converted below is ToolSearchToolResultBlock. WebSearchToolResultBlock, WebFetchToolResultBlock, CodeExecutionToolResultBlock, BashCodeExecutionToolResultBlock, and TextEditorCodeExecutionToolResultBlock still fall through and are dropped.

That breaks multi-turn use of HostedWebSearchTool and HostedCodeInterpreterTool. I ran a two-turn flow on this branch: GetResponseAsync with HostedWebSearchTool, then history.AddMessages(response), then a follow-up question. The second request sends this assistant message:

```json
{
  "role": "assistant",
  "content": [
    { "type": "server_tool_use", "id": "srvtoolu_ws1", "name": "web_search", "input": { "query": "weather sf" } },
    { "type": "text", "text": "It is sunny." }
  ]
}
````

The API rejects a server_tool_use in a non-final assistant message that lacks its matching *_tool_result (400, "web_search tool use with id … was found without a corresponding web_search_tool_result block"). Before this change the adapter dropped both blocks. That lost the search context but produced a valid request, so existing multi-turn callers now get 400s where they didn't before.

Could you make the round-trip symmetric by also converting the other server-tool result blocks to their *BlockParam counterparts with the same FromRawUnchecked(RawData) approach? Please also add a web-search multi-turn test next to the new tool-search round-trip tests, because no existing test covers server-tool blocks in request history.

jozkee: Addressed in the latest changes.  Good catch, thanks.

jozkee:
Will mark as draft until https://github.com/anthropics/anthropic-sdk-csharp/pull/242 is merged since it addresses the roundtrip concern and will keep this one focused on adding the tool.

PederHP:
Heads-up for the rebase once #242 lands: it now handles the server-tool round-trip via a single `ToServerToolBlockParam` helper (both adapters), and after review it also applies cache control to the replayed block and sets `hasHostedFiles` for a replayed `container_upload` — so the simplest resolution is to drop this PR's per-block `FromRawUnchecked` arms rather than merging the two sets. The `StreamedToolInput` helper here doesn't overlap (#242's `with { Input = input }` is in the MEAI streaming path, yours is in the aggregators), so that part is unaffected.

*Written by Claude Code, reviewed and posted by @PederHP.*

Issue #235, opened 2026-07-20
Optional request properties explicitly set to null are serialized as JSON null — beta-gated fields (e.g. context_management) fail with "Extra inputs are not permitted"


PederHP:
> [!NOTE]
> This issue was authored by Claude (Anthropic's Claude Fable 5, via Claude Code) and filed from my account. I have reviewed it and can vouch for the reproduction and the decompiled/source analysis. — @PederHP

## Summary

Optional request properties that are **explicitly initialized to `null`** are recorded in the params' raw body data and serialized onto the wire as JSON `null`. For beta-gated fields this makes the API reject the request outright. The most likely field to hit is `context_management`: a request carrying `"context_management": null` **without** the `context-management-2025-06-27` beta header fails with:

```
{"type":"error","error":{"type":"invalid_request_error","message":"context_management: Extra inputs are not permitted"}}
```

(Observed against `api.anthropic.com`, e.g. request id `req_011CdDasDM7eYX1G3JvMtMem`.)

I understand this is the documented contract — `MessageCreateParamsTest.OptionalNullableParamsSetToNullAreSetToNull_Works` asserts exactly this behavior, presumably for parity with the TypeScript SDK where `{foo: null}` and omitting `foo` are distinct. The problem is that **C# has no ergonomic way to conditionally omit a property** in an object initializer or `with` expression, so `Property = condition ? value : null` is the natural spelling of "only set this when enabled" — and it compiles cleanly, looks idiomatic, and produces a hard-to-diagnose 400 at runtime.

## Reproduction

```csharp
var parameters = new MessageCreateParams
{
    Model = "claude-opus-4-6",
    MaxTokens = 4096,
    Messages = [new() { Role = Role.User, Content = "hi" }],
    // Idiomatic-looking conditional config — this is the trap:
    ContextManagement = contextEditingEnabled ? myConfig : null,
};
// contextEditingEnabled == false  =>  body contains "context_management": null
// no beta header                  =>  400 "Extra inputs are not permitted"
```

The wire body can be confirmed without any HTTP call:

```csharp
parameters.RawBodyData.ContainsKey("context_management"); // true
```

Note that debugging this is doubly confusing because `JsonSerializer.Serialize(parameters)` goes through the class-level `JsonModelConverter`, which does **not** emit the null — only the actual HTTP path (`BodyContent()` serializing `RawBodyData` verbatim) does. So the natural diagnostic step ("serialize the params and look at them") hides the problem.

[...]

PR #236, opened 2026-07-20
fix(meai): add context-management beta header when raw template carries context_management

PederHP:
> [!NOTE]
> This PR was authored by Claude (Anthropic's Claude Fable 5, via Claude Code) and submitted from my account. I have reviewed the change and vouch for it. — @PederHP

Fixes #235 (the acute, adapter-path symptom; see the issue for the underlying set-to-null ergonomics discussion).

## Problem

A `ChatOptions.RawRepresentationFactory` template whose body carries `context_management` without the `context-management-2025-06-27` beta opt-in is rejected by the API:

```
context_management: Extra inputs are not permitted
```

This includes a property **explicitly initialized to null** — the params record it in the raw body data and serialize it as JSON `null` (the documented set-to-null contract), which is an easy trap since C# object initializers and `with` expressions cannot conditionally omit a property:

```csharp
RawRepresentationFactory = _ => new MessageCreateParams
{
    ...
    ContextManagement = contextEditingEnabled ? config : null, // 400 when disabled
};
```

## Change

In `AnthropicBetaClientExtensions.ToMessageCreateParams`, auto-add `context-management-2025-06-27` whenever the template's `RawBodyData` carries `context_management` — following the existing precedent where the adapter auto-adds `files-api-2025-04-14`, `skills-2025-10-02`, `code-execution-2025-08-25`, and `mcp-client-2025-11-20` based on request content. With the header present, an explicit `null` is accepted by the API and means "no context management", matching user intent.

## Tests

Two new tests in `AnthropicClientBetaExtensionsTests`:

- `GetResponseAsync_WithRawRepresentationFactory_ContextManagement_AddsBetaHeader` — config present, header auto-added, wire body verified.
- `GetResponseAsync_WithRawRepresentationFactory_NullContextManagement_AddsBetaHeader` — property explicitly null; the test also pins the current wire shape (`"context_management": null` in the body) and verifies the header makes it acceptable.

`./scripts/format`, `./scripts/lint`, and the full test suite against the Prism mock all pass (11,156 passed / 0 failed).

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
+++ b/src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
@@ -2566,4 +2566,142 @@ public class AnthropicClientBetaExtensionsTests : AnthropicClientExtensionsTests
         Assert.Equal("application/pdf", files[1].MediaType);
         Assert.Equal(2000, files[1].SizeInBytes);
     }
+
+    [Fact]
+    public async Task GetResponseAsync_WithRawRepresentationFactory_ContextManagement_AddsBetaHeader()
+    {
+        IEnumerable<string>? capturedBetaHeaders = null;
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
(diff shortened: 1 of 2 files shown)


sophie-ant (approved):
Thanks!

PR #242, opened 2026-08-06
fix(meai): round-trip server-tool blocks when replaying assistant turns

PederHP:
## What

The MEAI adapters surface server tool use/result blocks (tool search, web search/fetch, code execution, MCP) as `ToolCallContent`/`ToolResultContent` (or subclasses) whose fidelity lives in `RawRepresentation` — a **response**-typed block. The ChatMessage→params conversion only passes through `RawRepresentation is (Beta)ContentBlockParam`, so every replayed assistant turn **silently dropped these blocks**.

With interleaved thinking, the dropped blocks sit *between* signed thinking blocks, so the replayed turn no longer matches what the model generated and the API rejects it:

```
messages.N.content.M: `thinking` or `redacted_thinking` blocks in the latest assistant
message cannot be modified. These blocks must remain as they were in the original response.
```

This is hit in practice inside `FunctionInvokingChatClient`'s tool loop — e.g. the model runs the tool search tool, then calls a client-side function discovered via `defer_loading`; the mid-turn replay drops the `server_tool_use`/`tool_search_tool_result` blocks and the request 400s. Consumers can't work around it without inserting a custom pipeline stage between function invocation and the adapter. Tool-search results also need to be replayed for `defer_loading` tools to remain available to the model.

A second, related defect: the streamed `server_tool_use` content's `RawRepresentation` was the `content_block_start` block, whose `input` is empty (real input arrives via `input_json` deltas) — so even a manually converted block replayed as `"input": {}`. The beta adapter had the same problem for a streamed `mcp_tool_use`, which was surfaced straight from `content_block_start`: its `McpServerToolCallContent` had no `Arguments` at all.

## Changes

- `CreateMessageParams` (beta + GA): new switch arm mapping server-tool response blocks on `RawRepresentation` back onto the request param union. Response and param blocks are wire-compatible JSON, so they round-trip through the serializer verbatim (`ToServerToolBlockParam`). Cache control set on the MEAI content is applied to the replayed block, and a replayed `container_upload` keeps the files beta opt-in.
- `CreateStreamingToolCallContent` (beta + GA): rebuild the `server_tool_use` `RawRepresentation` with the input accumulated from `input_json` deltas, via `with` so `caller` and any field newer than the SDK survive.
- Beta only: a streamed `mcp_tool_use` is now created at `content_block_stop` with its deltas folded in through the shared `StreamedToolInput.WithMergedInput` (the helper the aggregators use). `advisor_tool_result` joins the replay arm, the cache-control arm and the streaming pass-through, where it was previously dropped.

## Behaviour changes

- Streaming, beta adapter: `McpServerToolCallContent` is yielded at `content_block_stop` rather than `content_block_start`, and now has `Arguments` populated. Previously it arrived earlier but empty.
- Streaming, beta adapter: an `advisor_tool_result` block now surfaces as an `AIContent` carrying the raw block instead of being skipped.
- No public API changes.

## Tests

Mirrored across the beta and GA adapter test classes unless noted:

- `GetResponseAsync_ReplayedServerToolBlocks_RoundTripToParams`: a replayed tool-search turn (thinking + `server_tool_use` + `tool_search_tool_result` + text) produces the exact original wire blocks.
- `GetResponseAsync_ReplayedServerToolBlocks_ApplyCacheControl`, `GetResponseAsync_ReplayedContainerUpload_RoundTripsToParams` (asserts the files beta header on the beta adapter).
- `GetStreamingResponseAsync_ServerToolUse_CapturesStreamedInputOnRawRepresentation`: streamed tool-search input lands on the raw block and `caller` is preserved.
- `GetStreamingResponseAsync_ServerToolTurn_ReplaysVerbatim`: a streamed tool-search turn replays as the exact blocks the model generated.
- Beta only: `GetStreamingResponseAsync_McpToolTurn_ReplaysVerbatim` and `GetStreamingResponseAsync_AdvisorToolTurn_ReplaysVerbatim`; both fail without the last commit.

Rebased onto `main` (retargeted from `next` per the new merge flow). Adapter and aggregator tests pass on `net8.0`; `scripts/format` and `scripts/lint` are clean. `net472` was not run locally (WSL), so that leg is left to CI.

https://claude.ai/code/session_01TNdy56BM26BYvTYvo1eSw9


src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
```diff
@@ -2654,140 +2568,154 @@ HostedFileContent file in fileClient.ListFilesAsync(
     }
 
     [Fact]
-    public async Task GetResponseAsync_WithRawRepresentationFactory_ContextManagement_AddsBetaHeader()
+    public async Task GetResponseAsync_ReplayedServerToolBlocks_RoundTripToParams()
```
sophie-ant, line 2795: Can we duplicate these tests for the non-beta extensions?

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -2243,13 +2083,28 @@ private static AIContent CreateStreamingToolCallContent(StreamingFunctionData fu
                 catch (JsonException) { }
             }
 
+            // The content_block_start block carries an empty input (streamed inputs arrive
+            // via input_json deltas); rebuild the RawRepresentation with the accumulated
+            // input so the block replays faithfully when the turn is sent back.
+            object? rawRepresentation = functionData.RawRepresentation;
+            if (rawRepresentation is ServerToolUseBlock rawBlock && input is not null)
+            {
+                rawRepresentation = new ServerToolUseBlock
```
sophie-ant: Can we preserve raw fields here?

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -986,6 +864,12 @@ out List<TextBlockParam>? systemMessages
                             contents.Add(rawContent);
                             break;
 
+                        case AIContent ac
+                            when ToServerToolBlockParam(ac.RawRepresentation)
+                                is { } serverToolBlock:
+                            contents.Add(serverToolBlock);
```
sophie-ant: I think we need a `WithCacheControlFrom` here

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -1796,30 +1613,55 @@ private static UsageDetails ToUsageDetails(
...
+        /// of the contents the adapter creates for server tool use and results) back onto
+        /// the request-side union so replayed assistant turns keep these blocks. Dropping
+        /// them modifies the replayed turn: the API rejects requests whose latest assistant
+        /// message no longer matches what the model generated (with interleaved thinking,
+        /// "`thinking` ... blocks in the latest assistant message cannot be modified"), and
+        /// tool-search results must be replayed for deferred tools to remain available.
+        /// Response and param blocks are wire-compatible JSON, so they round-trip through
+        /// the serializer verbatim.
+        /// </summary>
+        private static ContentBlockParam? ToServerToolBlockParam(object? rawRepresentation)
```
sophie-ant, line 1866: Does this work with streaming? Anything need updating here? https://github.com/anthropics/anthropic-sdk-csharp/blob/cb4cb123a0ca61f1ced8c48b969c5f5230eb41b0/src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs#L436

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1682,30 +1538,59 @@ private static UsageDetails ToUsageDetails(
...
+                case BetaServerToolUseBlock:
+                case BetaWebSearchToolResultBlock:
+                case BetaWebFetchToolResultBlock:
+                case BetaCodeExecutionToolResultBlock:
+                case BetaBashCodeExecutionToolResultBlock:
+                case BetaTextEditorCodeExecutionToolResultBlock:
+                case BetaToolSearchToolResultBlock:
+                case BetaMcpToolUseBlock:
+                case BetaMcpToolResultBlock:
+                case BetaContainerUploadBlock:
```
sophie-ant, line 1774: Since we look at files to determine if we should set the file beta header, is there any concern about the upload block here? Just want to make sure that's not going to break things. A unit test for this may be helpful

sophie-ant (changes requested):
Sorry about the delay reviewing! Had a few questions.

jozkee:
There's some overlap between https://github.com/anthropics/anthropic-sdk-csharp/pull/220, specifically https://github.com/anthropics/anthropic-sdk-csharp/pull/220/changes/f88a24c04aaf67cd44156c7fb93c683c7b136b49, and this PR.  Will mark mine as draft until this one gets merged.

src/Anthropic.Tests/AnthropicClientBetaExtensionsTests.cs
PederHP: Done. Mirrored including some new tests.

src/Anthropic/AnthropicClientExtensions.cs
PederHP: Yes. There is now `rawBlock with { Input = input }`.

src/Anthropic/AnthropicClientExtensions.cs
PederHP: Added including test.

src/Anthropic/AnthropicClientExtensions.cs
PederHP: Had Claude verify that it does: 

> result blocks arrive whole in `content_block_start`, and `server_tool_use` is rebuilt from the deltas. Added an end-to-end test that streams a tool-search turn and replays it verbatim (`…ServerToolTurn_ReplaysVerbatim`). Nothing needed at L436.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
PederHP: Ah yes missed this. New commit fixes this and there is a test to cover it.

PederHP:
> There's some overlap between #220, specifically [f88a24c](https://github.com/anthropics/anthropic-sdk-csharp/commit/f88a24c04aaf67cd44156c7fb93c683c7b136b49), and this PR. Will mark mine as draft until this one gets merged.

Note from Claude about a discrepancy between #220 and this one:

The streaming-input fix does not overlap. #220 adds `Core/StreamedToolInput.WithMergedInput` but only uses it in the SDK's own `MessageContentAggregator/BetaMessageContentAggregator` helpers, not in the MEAI streaming path. #242's `rawBlock with { Input = input }` in `CreateStreamingToolCallContent` stays needed either way. (After both land there'd be two ways of doing the same thing; `WithMergedInput` could later replace the` with` — a follow-up, not something for #242.)

PederHP:
@sophie-ant I hope it is ok that I do a friendly ping for when you have a moment. I addressed your earlier comments and just rebased onto current `main`.

PR #252, opened 2026-08-27
fix(bedrock): skip null-content events in streaming instead of crashing

latent-9:
`SseEventContentWrapper` crashes the Bedrock streaming read when the server sends an event-stream message without a `"bytes"` field (control events like `messageStart`, or `internal-error` payloads). The message is CRC-valid, so the parser accepts it and returns a null payload with `success=true`; the wrapper then calls `Encoding.UTF8.GetBytes(data!)`, which throws `ArgumentNullException`. Both the .NET and net472 read paths are affected.

Null payloads are now skipped and reading continues until a content event is found. Returning 0 was not an option: the SSE reader treats it as end-of-stream, which would silently truncate the response.

- Skip null-content events instead of crashing
- Return 0 only at true end-of-stream
- Two new tests (skip-then-read-next, clean EOF after a null event) using event-stream frames with valid CRCs

### DIFF
--- a/src/Anthropic.Bedrock/SseEventContentWrapper.cs
+++ b/src/Anthropic.Bedrock/SseEventContentWrapper.cs
@@ -79,22 +79,32 @@ internal class SseEventContentWrapper : HttpContent
                 return toCopy;
             }
 
-            var (data, success) = await AwsEventStreamHelpers
-                .ReadStreamMessage(_sourceStream, cancellationToken)
-                .ConfigureAwait(false);
-            if (!success)
+            while (true)
             {
-                return 0;
-            }
+                var (data, success) = await AwsEventStreamHelpers
+                    .ReadStreamMessage(_sourceStream, cancellationToken)
+                    .ConfigureAwait(false);
+                if (!success)
+                {
+                    return 0;
+                }
 
-            var encodedData = Encoding.UTF8.GetBytes(data!);
-            var bytesToCopy = Math.Min(encodedData.Length, buffer.Length);
-            encodedData.AsMemory(0, bytesToCopy).CopyTo(buffer);
-            if (bytesToCopy < encodedData.Length)
-            {
-                _remainder = encodedData.AsMemory(bytesToCopy);
+                // Messages without content (e.g. control events that carry no
+                // "bytes" field) yield a null payload; skip them and read on.
+                if (data == null)
+                {
+                    continue;
+                }
+
+                var encodedData = Encoding.UTF8.GetBytes(data);
+                var bytesToCopy = Math.Min(encodedData.Length, buffer.Length);
+                encodedData.AsMemory(0, bytesToCopy).CopyTo(buffer);
+                if (bytesToCopy < encodedData.Length)
+                {
+                    _remainder = encodedData.AsMemory(bytesToCopy);
+                }
+                return bytesToCopy;
             }
-            return bytesToCopy;
         }
 #else
         public override async Task<int> ReadAsync(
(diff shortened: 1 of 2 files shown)


latent-9:
Hi, this PR is ready for review. Could you approve the workflow run on this fork PR when you get a chance? Thanks.

feiiiiii5 (commented):
I ran your two tests against today's `main` and against your base. The picture has changed since you opened this: **#236 (`9b299f3bb2`, merged 2026-09-16) already fixed the crash, and it did it by moving the skip loop into the helper.** Everything below is reproducible; commands at the end.

**1. The `ArgumentNullException` is gone on main.** `AwsEventStreamHelpers.ReadStreamMessage` changed signature:

- at your base `4a01a3bc`: `Task<(string? Data, bool readData)> ReadStreamMessage(...)`
- on main `c4a1b81e`: `Task<string?> ReadStreamMessage(...)`, documented *"The next SSE event, or `null` at the end of the stream"*, and it now loops internally:

```csharp
while (true)
{
    var (data, readData) = await ReadFrame(source, cancellationToken).ConfigureAwait(false);
    if (!readData || data is not null)
    {
        return data;
    }
}
```

So a null-payload frame is skipped inside the helper, and `SseEventContentWrapper` no longer contains `Encoding.UTF8.GetBytes(data!)` — it is `GetBytes(data)` behind an `if (data is null) { return 0; }`.

Your own EOF test measures exactly that: `ReadAsync_NullContentEvent_AtEndOfStream_ReturnsZeroWithoutThrowing` **fails at `4a01a3bc`** and **passes on main unmodified**.

**2. As written, the PR no longer compiles against main.** Both `ReadAsync` overloads deconstructure `var (data, success) = await …ReadStreamMessage(…)`, and that method returns `string?` now.

**3. The skipping you are asking for is already main's behaviour**, so after a rebase the `SseEventContentWrapper` change has nothing left to do.

**4. One thing worth flagging regardless, because it would survive the rebase: the skip test fails on main for a fixture reason, not a product reason.** Its content frame carries the inner base64 payload `{"content":"hi"}`, which has no `type` field, and main's `Parse` returns `null` for a decoded event without one:

```csharp
// src/Anthropic.Bedrock/AwsEventStreamHelpers.cs:164
if (parsedEvent?["type"] is not JsonValue type || !type.TryGetValue(out string? eventType))
{
    return null;
}
```

So the loop skips *both* frames, reaches EOF and returns 0 — `Assert.True(read > 0, …)` fails even though the null-content skip you wanted did happen. I isolated this by changing one line and nothing else:

| inner payload in the fixture | product source | result on main `c4a1b81e` |
|---|---|---|
| `"{\"content\":\"hi\"}"` (as submitted) | unmodified | 1 failed, 1 passed |
| `"{\"type\":\"chunk\",\"content\":\"hi\"}"` | unmodified | **2 passed, 0 failed** |

**Suggestion:** close as superseded by #236 — the bug you found was real, and it is now fixed upstream. If you want the regression coverage to outlive this PR, the two tests are worth landing on their own with the fixture corrected and the wrapper change dropped; a null-content frame mid-stream is a shape worth pinning, since #236 made skipping it the documented contract rather than an accident.

Happy to be wrong about any of it if you are seeing something different against a real Bedrock stream — my runs are all against a `MemoryStream` of hand-built frames, which is what your test file does too.

Reproduce: in a worktree at each sha, add only `src/Anthropic.Tests/Bedrock/SseEventStreamNullContentTests.cs` and run `dotnet test src/Anthropic.Tests/Anthropic.Tests.csproj -f net8.0 --filter "FullyQualifiedName~SseEventStreamNullContentTests"`. .NET SDK 9.0.318, macOS arm64, `net8.0` only (`net472` needs mono here).

AI-assisted review; the measurements above are mine.

Issue #245, opened 2026-08-18
MEAI adapter: ToUsageDetails drops Usage.OutputTokensDetails.ThinkingTokens (UsageDetails.ReasoningTokenCount never populated)


deanxb:
The Microsoft.Extensions.AI adapter (AnthropicClientExtensions.AnthropicChatClient, verified on 12.40.0) never forwards Usage.OutputTokensDetails.ThinkingTokens. The response model carries the count (final Usage and streaming MessageDeltaUsage both), but ToUsageDetails forwards only input/output/cache-creation/cache-read and the two server-tool counters. So a consumer on the MEAI abstraction cannot meter reasoning tokens (for example, the OTel GenAI gen_ai.usage.reasoning.output_tokens attribute, added in semconv v1.41.0, is unreachable from this surface).

UsageDetails has had a first-class home for this since M.E.AI.Abstractions 10.1.x: ReasoningTokenCount (dotnet/extensions#7122), documented as counted within OutputTokenCount; the same semantics as the Anthropic field (thinking_tokens is documented as always ≤ output_tokens, with output_tokens remaining the inclusive billing total). The peer OpenAI adapter already populates it (OpenAIChatClient.FromOpenAIUsage: ReasoningTokenCount = tokenUsage.OutputTokenDetails?.ReasoningTokenCount), and the 12.40.0 package already depends on M.E.AI.Abstractions >= 10.5.1, so the fix looks like a one-liner: ReasoningTokenCount = usage.OutputTokensDetails?.ThinkingTokens plus the streaming-delta merge path.

Context: #221 is adjacent request-side reasoning evidence (the adapter's ChatOptions.Reasoning mapping).

Happy to PR this. Our downstream currently records the count as unreachable, with a pinned boundary test that alerts us if/when a release fixes it.

PR #261, opened 2026-09-12
fix(meai): populate UsageDetails.ReasoningTokenCount from output_tokens_details.thinking_tokens

serhiy-bzhezytskyy:
Fixes #245

## Summary

The Microsoft.Extensions.AI adapters never forward `Usage.OutputTokensDetails.ThinkingTokens`: `ToUsageDetails` (`AnthropicClientExtensions.cs`, and its beta twin) passes input, output, cache-creation, cache-read and server-tool counts into the shared overload and stops there, so a consumer on the `IChatClient` surface cannot meter reasoning tokens even though the response model carries them on both the final `usage` and the cumulative `message_delta` usage.

`UsageDetails.ReasoningTokenCount` exists in the pinned `Microsoft.Extensions.AI.Abstractions` 10.5.1 and is documented as counted within `OutputTokenCount` — the same relationship the API documents for `thinking_tokens` and `output_tokens`, so the mapping is a direct assignment with no arithmetic.

## Change

- `ToUsageDetails(Usage)` / `ToUsageDetails(BetaUsage)` pass `usage.OutputTokensDetails?.ThinkingTokens`; the shared overload takes it as `long? thinkingTokens` and sets `ReasoningTokenCount`.
- Streaming: the `message_start` value is remembered and the `message_delta` path uses `deltaUsage.OutputTokensDetails?.ThinkingTokens ?? startThinkingTokens`, the same fallback the cache-token fields already use so a terminal delta that omits the details does not wipe the count.
- No behaviour change when `output_tokens_details` is absent: `ReasoningTokenCount` stays `null`, as before.

## Tests

Two tests in `AnthropicClientExtensionsTestsBase`, so they run against both adapters: a non-streaming response with `"output_tokens_details": {"thinking_tokens": 30}` → `ReasoningTokenCount == 30`; a stream whose `message_delta` carries `thinking_tokens: 25` → the last `UsageContent` reports 25. With the adapter change reverted all four fail with `Actual: null`. `AnthropicClientExtensionsTests` 249 and `AnthropicClientBetaExtensionsTests` 261 pass on `net8.0`; `csharpier` and `dotnet format` clean.

Assisted by Claude Code.

### DIFF
--- a/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
+++ b/src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
@@ -6570,6 +6570,120 @@ public abstract class AnthropicClientExtensionsTestsBase
     }
 
     [Fact]
+    public async Task GetResponseAsync_WithThinkingTokens_PopulatesReasoningTokenCount()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "Test with thinking"
+                    }]
+                }],
+                "max_tokens": 1024
+            }
+            """,
+            actualResponse: """
+            {
+                "id": "msg_thinking_01",
+                "type": "message",
+                "role": "assistant",
+                "model": "claude-haiku-4-5",
+                "content": [{
+                    "type": "text",
+                    "text": "Response"
+                }],
+                "stop_reason": "end_turn",
+                "usage": {
+                    "input_tokens": 100,
+                    "output_tokens": 40,
+                    "output_tokens_details": {
+                        "thinking_tokens": 30
+                    }
+                }
+            }
+            """
+        );
+
+        IChatClient chatClient = CreateChatClient(handler, "claude-haiku-4-5");
+
+        ChatResponse response = await chatClient.GetResponseAsync(
+            "Test with thinking",
+            new(),
+            TestContext.Current.CancellationToken
+        );
+
+        Assert.NotNull(response.Usage);
+        Assert.Equal(40, response.Usage.OutputTokenCount);
+        Assert.Equal(30, response.Usage.ReasoningTokenCount);
+    }
+
+    [Fact]
+    public async Task GetStreamingResponseAsync_ThinkingTokensFromDelta_PopulateReasoningTokenCount()
+    {
+        VerbatimHttpHandler handler = new(
+            expectedRequest: """
+            {
+                "max_tokens": 1024,
+                "model": "claude-haiku-4-5",
+                "messages": [{
+                    "role": "user",
+                    "content": [{
+                        "type": "text",
+                        "text": "hello"
+                    }]
+                }],
+                "stream": true
+            }
+            """,
+            actualResponse: """
+            event: message_start
+            data: {"type":"message_start","message":{"id":"msg_01","type":"message","role":"assistant","model":"claude-haiku-4-5","content":[],"stop_reason":null,"stop_sequence":null,"usage":{"input_tokens":8,"cache_creation_input_tokens":0,"cache_read_input_tokens":0,"output_tokens":1,"output_tokens_details":{"thinking_tokens":0}}}}
+
+            event: content_block_start
+            data: {"type":"content_block_start","index":0,"content_block":{"type":"text","text":""}}
+
+            event: content_block_delta
+            data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"Hello!"}}
+
+            event: content_block_stop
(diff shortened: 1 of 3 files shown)


feiiiiii5 (commented):
**Disclosure first:** I have three open PRs in this repo (#268, #270, #271), and #270/#271 both touch `AnthropicBetaClientExtensions.cs`. I checked for overlap before reviewing — `git merge-tree --write-tree` reports a clean merge between this branch and all three of mine, so nothing here is me defending a conflict. Weigh the rest accordingly.

I read the diff against `main` @ `c4a1b81e` rather than running it; the two points below are deductive from code that is visible in the diff, and I have flagged where I did not execute.

### The mapping is right, including the part that could have gone wrong

Two things I verified because they are the ways this kind of change usually breaks:

- **No double-count.** In the shared overload, `usageDetails.TotalTokenCount = NullableSum(usageDetails.InputTokenCount, usageDetails.OutputTokenCount)` — it does not include `ReasoningTokenCount`. So treating `thinking_tokens` as a subset of `output_tokens` (as the comment says, and as `UsageDetails.ReasoningTokenCount` is documented) leaves the total correct. A direct assignment really is the right shape here.
- **No TFM gap.** `Microsoft.Extensions.AI.Abstractions` `10.5.1` is referenced unconditionally at `src/Anthropic/Anthropic.csproj:25`, while `src/Directory.Build.props:27` multi-targets `net8.0;net9.0;netstandard2.0`. The reference is not inside a `'$(TargetFramework)' == 'netstandard2.0'` ItemGroup, so `ReasoningTokenCount` is available on all three targets and the `net472` test run (which exercises the `netstandard2.0` build) covers it too.

Symmetry between the stable and beta adapters is exact — same five hunks, same ordering — which is what CLAUDE.md asks for.

### The gap: the new `?? startThinkingTokens` fallback is never exercised

This is the one change I would ask for. Neither new test can reach the right-hand operand:

- `GetResponseAsync_WithThinkingTokens_PopulatesReasoningTokenCount` is non-streaming, so there is no start/delta interaction at all.
- `GetStreamingResponseAsync_ThinkingTokensFromDelta_PopulateReasoningTokenCount` puts `"thinking_tokens": 25` **in the `message_delta`**, so `deltaUsage.OutputTokensDetails?.ThinkingTokens` is non-null and the `??` short-circuits. `startThinkingTokens` is written by `message_start` (as `0`) and then never read.

So deleting `?? startThinkingTokens` from both adapters would leave both tests green — the fallback is the branch that stops a terminal delta from wiping the count, and it is currently the only part of the change with no test that can fail. (Reasoned from the fixtures in the diff; I did not execute the mutation.)

The file already contains the template, immediately below the insertion point: `GetStreamingResponseAsync_DeltaWithoutCacheTokens_PreservesStartCacheTokens` (`AnthropicClientExtensionsTestsBase.cs:6762` at `main`), which is exactly this test for the cache-token fields this one mirrors. A twin — `message_start` carrying a non-zero `thinking_tokens` (say 7), then a `message_delta` whose `usage` omits `output_tokens_details` entirely, asserting `ReasoningTokenCount == 7` — would cover the branch and match the existing convention. The non-zero start value matters: with `thinking_tokens: 0` at `message_start`, a regression to "wipe it" is not distinguishable from "never set" by an `Assert.Equal(0, ...)`.

### Base branch

This targets `next`. `CLAUDE.md:10` is explicit that it should not:

> PRs target the default branch (`main` on the public repo). `next` belongs to release automation — don't open PRs against it or push to it; a commit that lands there outside the release flow breaks the sync.

Retargeting looks mechanical rather than a rebase with real work in it: `git merge-tree --write-tree --name-only origin/main <head>` against current `main` (`c4a1b81e`) returns a tree with no conflicted paths, so the three files apply cleanly. I hit the same thing on my own PRs this week and the switch was just a base change plus re-stating the base sha in the description.

For what it is worth, this may simply predate the rule: the string "PRs target the default branch" does not appear in `CLAUDE.md` at this branch's base commit (`4e952bd2`), while it is at `CLAUDE.md:10` on current `main`. The `next` snapshot this was branched from carries older contributor docs — one more reason to re-point at `main` before review rather than after.

---
*Read against `main` @ `c4a1b81e` and PR head @ `4e952bd2`; merge-tree results are from those commits. No test run performed, so the pass counts in the description are the author's, not mine.*

serhiy-bzhezytskyy:
Thanks for the review, and for checking the overlap with your own PRs first.

Fixed the test gap: added `GetStreamingResponseAsync_DeltaWithoutThinkingTokens_PreservesStartThinkingTokens`, the twin of the existing cache-token test for the same shape -- non-zero `thinking_tokens` at `message_start`, `message_delta` omits `output_tokens_details` entirely, asserts the start value survives. `0f42004b`.

This branched before CLAUDE.md's base-branch rule flipped. Retargeting to `main`; `git merge-tree` confirms a clean apply, no rebase needed.

Issue #263, opened 2026-09-16
AsIChatClient: ChatToolMode.RequireSpecific(name) silently discards the tool name and sends tool_choice: {"type":"any"}


lucachitayat:
`ChatToolMode.RequireSpecific(name)` silently discards the tool name and sends `tool_choice: {"type":"any"}`.

## Environment

| | |
|---|---|
| `Anthropic` (NuGet) | **12.40.0** (also reproduced at **12.19.0**) |
| `Microsoft.Extensions.AI` | **10.9.0** |
| Target framework | `net10.0` |
| .NET SDK | 10.0.400 |
| OS | Windows 11, 10.0.26200 |
| Surface | `Anthropic.Models.Messages` (stable) **and** `Anthropic.Beta` — see *Mechanism* |

Still present on `main` at `4a48803` (latest release `Anthropic-v12.48.0`).

> **Provenance of the payloads below.** Every wire payload in this report was captured from an
> offline harness — the real adapter and the real serializer, over a stubbed
> `HttpMessageHandler` — not from a live API call. That is sufficient here because the defect is
> in what the SDK *sends*, so the request body is the whole of the evidence. Source citations are
> to `main` at the commit above.

## Expected vs actual

Given two registered tools and `ChatToolMode.RequireSpecific("get_weather")`:

- **Expected:** `"tool_choice": {"type": "tool", "name": "get_weather"}`
- **Actual:** `"tool_choice": {"type": "any"}` — no `name` member at all

The request succeeds. There is no exception, no warning, and no log line. Register several tools,
demand one of them by name, and the model may call any of the others.

The two `ChatToolMode` cases are indistinguishable on the wire: `RequireAny` and
`RequireSpecific("get_weather")` produce byte-identical `tool_choice` objects.

## Observed wire payload

Both captured at 12.40.0.

```jsonc
// ChatToolMode.RequireAny
"tool_choice": { "type": "any" }

// ChatToolMode.RequireSpecific("get_weather")  <-- the defect
"tool_choice": { "type": "any" }
```

For contrast, the payload once a decorator pre-sets `ToolChoice` itself (see *Workaround*):

```jsonc
"tool_choice": { "type": "tool", "name": "get_weather" }
```

## Minimal reproduction

Self-contained apart from the capturing transport, which is ordinary `HttpMessageHandler`
plumbing. This is the exact call shape I use to pin it; the assertion reads `tool_choice.type` and
[...]

PR #265, opened 2026-09-16
fix(meai): honour RequiredChatToolMode.RequiredFunctionName as tool_choice.tool

lucachitayat:
Fixes #263.

## What was wrong

In `Microsoft.Extensions.AI`, *require any tool* and *require one specific tool* are the **same type** — `RequiredChatToolMode` — differing only by a nullable `RequiredFunctionName`. Both adapters discriminated on the runtime type alone, so the name was dropped on the floor.

```csharp
: toolMode is RequiredChatToolMode ? new ToolChoiceAny() { ... }
```

Before / after, for `ChatToolMode.RequireSpecific("get_weather")` with two tools registered:

```jsonc
// before -- byte-identical to ChatToolMode.RequireAny
"tool_choice": { "type": "any" }

// after
"tool_choice": { "type": "tool", "name": "get_weather" }
```

The call succeeded either way. No exception, no warning, no log line — the caller demanded one tool and the model was free to call any of the others.

Per #263 (**MEASURED**, compiled assembly metadata): `RequiredChatToolMode.RequiredFunctionName`'s getter has zero occurrences anywhere in the shipped `Anthropic` assembly, across all three TFM groups, in both 12.19.0 and 12.40.0 — the property is never read by either adapter.

## What changed

One new arm in the `tool_choice` mapping in each of the two adapters, matching on the property rather than the type:

```csharp
: toolMode is RequiredChatToolMode { RequiredFunctionName: { } name }
    ? new ToolChoiceTool() { Name = name, DisableParallelToolUse = !options.AllowMultipleToolCalls }
: toolMode is RequiredChatToolMode
    ? new ToolChoiceAny() { ... }   // unchanged
```

Applied to both copies per the "a fix usually needs applying to every copy" note in `CLAUDE.md`: `AnthropicClientExtensions` (stable, `ToolChoiceTool`) and `AnthropicBetaClientExtensions` (beta, `BetaToolChoiceTool`). Both surfaces had the defect. A repo-wide sweep found no third copy: streaming shares the same private `GetMessageCreateParams` in each adapter, and the five platform packages plus `Anthropic.Mcp` contain no `ChatOptions`/`ToolMode`/`ToolChoice` references at all.

`DisableParallelToolUse` is carried over from the `ToolChoiceAny` arm unchanged. Verified at the wire on both surfaces: `null` → field absent, `true` → `false`, `false` → `true`.

## Behaviour changes, both intended

1. **A caller who passed `RequireSpecific` was silently getting `"any"` and will now get `"tool"`.** That is the point of the fix, but a request that previously let the model choose will now force one tool.
2. **A `RequiredFunctionName` naming a tool that is not in `options.Tools` now goes to the API as-is**, where before it was silently replaced with `"any"` and the request succeeded. Nothing validates the name — not `ToolChoiceTool.Validate()`, not `MessageCreateParams`, not MEAI. Wire bytes confirmed; the resulting 400 is inferred, not measured. This matches the "map inputs as-is and let the API reject what it doesn't support" line in `CLAUDE.md`, and the previous behaviour was worse, but it is a real change for anyone who was passing a bogus name.

Worth noting what is *not* affected: `any` and `tool` have identical support profiles in the docs — wherever forced tool use is rejected (manual extended thinking; Fable 5.1 / Mythos 5.1), both are rejected together. So there is no configuration where the old `"any"` worked and the new `"tool"` fails.

## What gives confidence

Two tests in `AnthropicClientExtensionsTestsBase`, so both run against **both** adapters, asserting the exact wire request through `VerbatimHttpHandler` (`JsonNode.DeepEquals`, so an extra *or* missing member fails).

`GetResponseAsync_WithToolModeRequireSpecific` registers two tools and requests **the second one**. That detail is load-bearing, and I got it wrong the first time: the original version requested `get_weather`, which is `Tools[0]`, so an implementation reading the first registered tool's name instead of `RequiredFunctionName` passed it. Measured, with exactly such an implementation injected into both adapters:

| | original test (requests `Tools[0]`) | current test (requests `Tools[1]`) |
|---|---|---|
| wrong impl (reads `Tools[0].Name`) | **passes 2/2** — blind | **fails 2/2** |
| this PR's impl | passes 2/2 | passes 2/2 |

`GetResponseAsync_WithToolModeRequireSpecificAndParallelToolCallsDisabled` pins `AllowMultipleToolCalls = false` → `"disable_parallel_tool_use": true` on the new arm, which the first test leaves unexercised.

- **Fails before:** 2 failed (one per adapter), both with `"tool_choice":{"type":"any"}` in the actual body.
- **Full suite:** `upstream/next` 126 failed / 14970 passed / 15127 total; this branch 126 failed / **14974** passed / 15131 total. Identical failure set — all 126 are generated `*ServiceTest` methods hitting `Connection refused` on the port-4010 mock server, which I did not start. +4 passing = 2 new tests × 2 adapters.
- `./scripts/build` clean on all three TFMs including `netstandard2.0` (0 warnings, warnings-as-errors); `./scripts/lint` clean.

Not run locally: the `net472` execution, which needs Windows. The `netstandard2.0` build compiles clean and the change is plain pattern matching with no TFM-sensitive API, but CI is the source of truth.

## Notes

- No user-derived `ChatToolMode` can exist, so the arm ordering cannot be bypassed from outside: `RequiredChatToolMode`, `AutoChatToolMode` and `NoneChatToolMode` are all `sealed`, and base `ChatToolMode`'s only constructor is `private protected`.
- Pre-existing, for whoever reads this next: `ChatToolMode.RequireSpecific(null!)` does not throw — MEAI guards its `Throw.IfNullOrWhitespace` behind a null check — so it produces a `RequiredChatToolMode` indistinguishable from `RequireAny` and correctly falls through to the `"any"` arm. `""` and `"   "` do throw.
- A fail-loudly alternative (throwing on a `RequiredFunctionName` the adapter cannot honour) would also beat discarding it, but the SDK can honour it, so mapping it is better.
- Companion PR #266 touches a different region of the same file; they should merge in either order without conflicting.
- Nothing to document outside this repo — this makes an already-documented `ChatOptions` knob work as its name says.


src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -1655,6 +1655,14 @@ toolMode is AutoChatToolMode
                                 DisableParallelToolUse = !options.AllowMultipleToolCalls,
                             }
                         : toolMode is NoneChatToolMode ? new ToolChoiceNone()
+                        // RequiredChatToolMode covers both "any tool" and "this tool";
+                        // only RequiredFunctionName tells the two apart.
+                        : toolMode is RequiredChatToolMode { RequiredFunctionName: { } name }
```
lucachitayat, line 1660: Arm ordering is load-bearing: the property pattern has to come **before** the bare `toolMode is RequiredChatToolMode` arm below, since every `RequireSpecific` also matches the bare test. Swapping them silently restores the old behaviour, and no existing test would catch it — which is why the new test registers two tools rather than one.

`DisableParallelToolUse` is copied verbatim from the `ToolChoiceAny` arm rather than reformulated, so `AllowMultipleToolCalls` keeps its exact current semantics on this path, `null` included.

src/Anthropic/Services/Beta/Messages/AnthropicBetaClientExtensions.cs
```diff
@@ -1524,6 +1524,14 @@ toolMode is AutoChatToolMode
                                 DisableParallelToolUse = !options.AllowMultipleToolCalls,
                             }
                         : toolMode is NoneChatToolMode ? new BetaToolChoiceNone()
+                        // RequiredChatToolMode covers both "any tool" and "this tool";
+                        // only RequiredFunctionName tells the two apart.
+                        : toolMode is RequiredChatToolMode { RequiredFunctionName: { } name }
```
lucachitayat, line 1529: Same change against `BetaToolChoiceTool`. The beta surface had the defect independently — unlike the schema transform, this mapping is genuinely duplicated rather than shared, per the "a fix usually needs applying to every copy" note in `CLAUDE.md`. I checked for further copies: these two are the only sites that map `ChatToolMode` to a `tool_choice`.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -1967,6 +1967,106 @@ public async Task GetResponseAsync_WithToolModeRequired()
         Assert.NotNull(response);
     }
 
+    [Fact]
+    public async Task GetResponseAsync_WithToolModeRequireSpecific()
+    {
```
lucachitayat, line 1972: Placed in the shared base rather than a derived class so it runs against both adapters — it fails twice before the fix and passes twice after.

Two tools are registered on purpose. With one tool, `{"type":"any"}` and `{"type":"tool","name":"get_weather"}` request the same behaviour from the model, so the test would pin the wire shape without pinning anything that matters. With two, the old output genuinely permits a tool the caller did not ask for.

`VerbatimHttpHandler` compares with `JsonNode.DeepEquals`, so this also fails on a `name` that is present but misplaced, not just on one that is missing.

lucachitayat (commented):
Self-review notes on the two things a reviewer would most reasonably push back on.

sophie-ant (dismissed):
LGTM. Could you retarget against `main`? We recently changed how the different branches work; we're going to update CLAUDE.md shortly. Sorry about that!

sophie-ant (approved):
I can just change the base branch myself 😼

lucachitayat:
> I can just change the base branch myself 😼

Thanks so much @sophie-ant. I actually got this in before leaving on a vacation. Happy to see it made it in!

Issue #264, opened 2026-09-16
AsIChatClient: union-typed schema nodes ("type": ["string","null"]) lose format / properties / required into description


lucachitayat:
A union-typed schema node (`"type": ["string","null"]`) loses its type-specific keywords — `format`, and on the object branch `properties`/`required`/`additionalProperties` — which are re-stringified into `description`.

## Environment

| | |
|---|---|
| `Anthropic` (NuGet) | **12.40.0** |
| `Microsoft.Extensions.AI` | **10.9.0** |
| Target framework | `net10.0` |
| .NET SDK | 10.0.400 |
| OS | Windows 11, 10.0.26200 |
| Models used for the live enforcement check | `claude-sonnet-4-6`, `claude-haiku-4-5` |

Still present on `main` at `4a48803` (latest release `Anthropic-v12.48.0`). Source citations below
are to that commit.

> **Provenance, stated up front because it varies by claim.**
> - Every **request** payload below was captured from an offline harness — the real adapter and
>   the real serializer over a stubbed `HttpMessageHandler` — not from a live API call. That is
>   sufficient for the defect itself, which is about what the SDK *sends*.
> - The one claim that needed the **live API** is *"the platform actually enforces `format` on a
>   union-typed node"*, measured on 2026-09-14 and reported in its own section below.
> - Property names in the samples are neutral placeholders; the JSON *structure* is verbatim.

## Expected vs actual

A property whose `type` is a JSON **array** — the shape `System.Text.Json` emits for
`Nullable<T>` and for a nullable reference type — loses its type-specific JSON Schema keywords.
Its non-nullable sibling of the same type keeps them.

- **Expected:** `"endsOn": { "type": ["string","null"], "format": "date" }`, or any equivalent
  shape that preserves the constraint
- **Actual:** `"endsOn": { "type": ["string","null"], "description": "…\n\n{format: \"date\"}" }`
  — **no `format` member at all**

`date` is on the [documented supported-format list](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations),
and the documented SDK transformation **filters string formats to the supported list** rather than
stripping them — step 4 of
[§ How SDK transformation works](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#how-sdk-transformation-works),
which names the C# SDK explicitly. The constraints that *are* meant to be stripped and
[...]

PR #266, opened 2026-09-16
fix(meai): keep type-specific keywords on union-typed schema nodes

lucachitayat:
Fixes #264.

## What was wrong

The schema transform selects a per-type keyword allowlist by sniffing the node's `type`, and the sniff tested for a JSON *value*:

```csharp
string? type =
    schemaObj.TryGetPropertyValue("type", out JsonNode? typeNode)
    && typeNode is JsonValue          // a JsonArray fails here
        ? typeNode.GetValue<string>()
        : null;
```

A union type — `"type": ["string","null"]`, the shape `System.Text.Json` emits for `Nullable<T>` and for a nullable reference type — is a JSON *array*, so it read as typeless, fell through to `s_supportedBaseSchemaProperties`, and every type-specific keyword on the node was removed and re-stringified into `description`.

Before / after, on the same generated schema, where each nullable member differs from its non-nullable sibling **only** in nullability:

```jsonc
// before
"startsOn": { "type": "string", "format": "date" },                       // control, survives
"endsOn":   { "type": ["string","null"],
              "description": "An end date, or null.\n\n{format: \"date\"}" },
"tags":     { "type": ["array","null"],
              "description": "{items: {\"type\":\"string\"}, minItems: 3}" },
"nested":   { "type": ["object","null"],
              "description": "{properties: {\"id\":{\"type\":\"string\"}}, required: [\"id\"], additionalProperties: false}" }

// after
"startsOn": { "type": "string", "format": "date" },                       // unchanged
"endsOn":   { "type": ["string","null"], "description": "An end date, or null.", "format": "date" },
"tags":     { "type": ["array","null"], "items": { "type": "string" }, "description": "{minItems: 3}" },
"nested":   { "type": ["object","null"], "properties": { "id": { "type": "string" } },
              "required": ["id"], "additionalProperties": false }
```

Two things make this worth fixing rather than documenting:

- `date` is on the [documented supported-format list](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations), and the documented transformation **filters string formats to the supported list** rather than stripping them — step 4 of [§ How SDK transformation works](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#how-sdk-transformation-works), which names the C# SDK explicitly. The constraints that *are* meant to be stripped and re-described — `minimum`, `maximum`, `multipleOf`, `minLength`, `maxLength` — are listed separately under *Not supported*. So the transform was applying the strip-and-describe treatment to a keyword the docs say is filtered-and-kept.
- The object case is the quiet one. The root keeps its own `properties`/`required`, so `output_config` still goes out and the response still deserializes — the nested member has simply become an unconstrained object. Nothing goes red.

Type arrays are documented, not merely tolerated: [§ Schema complexity limits](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#schema-complexity-limits) meters *"parameters that use `anyOf` or type arrays (for example, `"type": ["string", "null"]`)"* at 16 per request, and the *Not supported* list does not mention them. #264 has the live-API measurement showing the platform honours `format` on a widened type array (6/6 on `claude-sonnet-4-6` and `claude-haiku-4-5`, with a within-call control).

## What changed

**Commit 1 — select the allowlist from every type the node names.**

```csharp
HashSet<string> supported;
if (unionTypes is null)
{
    supported = GetSupportedSchemaProperties(type);
}
else
{
    supported = new(s_supportedBaseSchemaProperties, StringComparer.Ordinal);
    foreach (string memberType in unionTypes)
    {
        supported.UnionWith(GetSupportedSchemaProperties(memberType));
    }
}
```

The `type switch` that was inline is now `GetSupportedSchemaProperties`, unchanged in content, so it can be called per member. The two type-specific pre-checks (`type == "string"` for `format`, `type == "array"` for `minItems`) now test membership, which is why `phone` and `tags` above still get their *unsupported* values stringified — a union node is treated on the same terms as its scalar counterpart in both directions, not just the permissive one.

**Commit 2 — validate keyword values before keeping them.** Widening those pre-checks to union nodes exposed two latent problems in them, one newly reachable. I found these in review after opening the PR; measured against `upstream/next`:

| union node input | `upstream/next` | after commit 1 | now |
|---|---|---|---|
| `format: 3` | `"description":"{format: 3}"` | **throws `InvalidOperationException`** out of `GetResponseAsync` | back to `"{format: 3}"` |
| `format: {"a":1}` | stringified | **throws** | stringified |
| `format: null` | stringified | `"format":null` on the wire | stringified |
| `minItems: "5"` | stringified | `"minItems":"5"` on the wire | stringified |
| `minItems: 2147483648` | stringified | emitted verbatim, defeating the code's own 0/1 rule | stringified |

Cause: the `format` check read the value with an unguarded `GetValue<string>()`, and both checks failed *open* on a malformed value — skipping the pre-check, after which the allowlist kept the keyword. Both are now inverted to keep the keyword only when its value is well formed and supported. This also fixes the same `GetValue<string>()` throw on a **scalar** `"string"` node, which predates this PR (`{"type":"string","format":3}` threw on `upstream/next` too).

Commit 2 also drops the per-node `HashSet` allocation on the scalar path — union members are collected only for an actual type array — and scopes the cross-SDK parity remark in the doc comment (see *Divergence* below).

This is the only copy of this transform: `AnthropicBetaClientExtensions` consumes `AnthropicClientExtensions.JsonSchemaTransformCache` rather than defining its own, so the beta surface is fixed by the same change. Both the `ResponseFormat` and the `AIFunctionDeclaration` tool paths go through this cache — path equivalence is already pinned by the existing `GetResponseAsync_ToolDeclarationSchema_AllTransformationsApplied` / `…ResponseFormatSchema_AllTransformationsApplied` pair, so the new tests below cover the `ResponseFormat` path only rather than duplicating that.

For a scalar **string** `type` the allowlist behaviour is byte-identical; `unionTypes` stays null and resolves to the same static set as before.

## Divergence from the other SDKs, stated deliberately

`CLAUDE.md` makes cross-SDK matching the default, so this needs saying plainly: TypeScript's `transformJSONSchema` compares `type` with `===` against `'object'`/`'string'`/`'array'`, so a type array matches no branch and it behaves exactly like the old C# code; Python's `transform_schema` ends in `assert_never(type_)` and **raises** on a list-valued `type`. The union-of-allowlists behaviour here is a deliberate divergence from both, and the stale "same whitelist approach as the TypeScript and Python SDKs" remark in the doc comment is now scoped to say so.

[anthropic-sdk-python#1892](https://github.com/anthropics/anthropic-sdk-python/pull/1892) fixes the same root cause the other way, by rewriting `{"type": ["string","null"], "format": "date"}` into `{"anyOf": [{"type": "string", "format": "date"}, {"type": "null"}]}`. I went narrower here because `CLAUDE.md` says to **map inputs as-is** rather than rewrite what the caller built, and because the type array is a documented, budgeted construct that needs no rewriting to be enforced — the docs count `anyOf` and type arrays identically against the same limit of 16, so the expansion buys nothing. It also avoids recursively re-transforming synthesized branches. That PR is still open (verified), so there is no merged cross-SDK shape to match yet; happy to switch to the `anyOf` form if you'd rather have parity now — only the test expectations would change.

## Other behaviour changes worth calling out

- **Constraints now bind.** Schemas that previously went out with a stripped union node will now carry the keyword, so the model is genuinely constrained where before it was only nudged via the description. That is the intent, but it changes model output for callers who adapted to the looser schema, and it moves the wire shape and therefore prompt-cache prefixes.
- **Key order inside a generated `description` flips for union nodes** carrying several stripped keywords, because `format`/`minItems` are now hoisted by the pre-checks instead of being picked up in property order by the generic loop. It makes unions consistent with scalars. Untested and out of the stated scope; flagging rather than hiding it.
- **`Dictionary<string,T>?` now emits `additionalProperties: {…}`** where it previously stringified it, and the docs list a non-`false` `additionalProperties` as a 400. Non-nullable `Dictionary<string,T>` already emits exactly this today, so the change makes the nullable case match an already-broken sibling rather than breaking new ground — but it is a case that previously succeeded. Happy to strip non-`false` `additionalProperties` in this PR if you'd prefer; I left it out because it changes the non-nullable path too, which is a separate logical change.

## Adjacent gaps I did not fix

Found while auditing the allowlist; all pre-existing and cross-SDK-consistent, so out of scope for this PR, but they are the same class of bug:

- `default` — [documented supported](https://platform.claude.com/docs/en/build-with-claude/structured-outputs#json-schema-limitations) ("`default` property for all supported types"), absent from the base allowlist, so stripped into `description`. The existing test at `AnthropicClientExtensionsTestsBase.cs` enshrines that.
- `definitions` and `$def` — the docs list "`$ref`, `$def`, and `definitions`" as supported; the allowlist has only `$ref` and `$defs`, so a draft-07 schema loses `definitions` and is left with dangling `$ref`s.
- `pattern` — not in *Not supported*, and the docs carry a dedicated "Pattern support (regex)" section; stripped all the same.

Also: the docs scope C# transformation to schemas *"derived from a native type (`Create<T>()`)"*. This adapter also transforms **caller-supplied** schemas passed through `ChatOptions.ResponseFormat`, which is outside what the docs describe — see the "obvious workaround does not work" section of #264.

## What gives confidence

Two tests in `AnthropicClientExtensionsTestsBase`, so both run against both adapters, asserting the exact wire request through `VerbatimHttpHandler`.

`GetResponseAsync_ResponseFormatSchema_UnionTypeKeepsTypeSpecificKeywords` carries a non-nullable `startsOn` control in the same schema, so it distinguishes "nullability breaks `format`" from "the transform always strips `format`". Members chosen to cover each allowlist that exists: `["string","null"]` for `format`, `["array","null"]` for `items`, `["object","null"]` for `properties`/`required`/`additionalProperties`.

`GetResponseAsync_ResponseFormatSchema_MalformedKeywordValuesAreStripped` pins the commit-2 table above.

| | union test | malformed test |
|---|---|---|
| `upstream/next` | fails 2/2 | fails 2/2 (the scalar `format: 3` crash) |
| after commit 1 | passes 2/2 | fails 2/2 |
| now | passes 2/2 | passes 2/2 |

- **Full suite:** `upstream/next` 126 failed / 14970 passed / 15127 total; this branch 126 failed / **14974** passed / 15131 total. Identical failure set — all 126 are generated `*ServiceTest` methods hitting `Connection refused` on the port-4010 mock server, which I did not start. +4 passing = 2 new tests × 2 adapters.
- `GetResponseAsync_ResponseFormatSchema_AllTransformationsApplied` and its tool-path twin still pass; they pin the scalar-type behaviour this change must not move.
- `./scripts/build` clean on all three TFMs including `netstandard2.0` (0 warnings, warnings-as-errors); `./scripts/lint` clean.

Not run locally: the `net472` execution, which needs Windows. The `netstandard2.0` build compiles clean, but CI is the source of truth.

## Notes

- Companion PR #265 touches a different region of the same file; they should merge in either order without conflicting.
- `Anthropic.Helpers.SchemaInference`, which backs `client.Messages.Create<T>()`, has its own and quite different union handling — it collapses `["string","null"]` to `"string"` and drops the member from `required`. Not this defect, not touched here, but it means the two public surfaces emit different schemas for the same C# type. Pre-existing; worth its own issue.
- #264 also notes `$schema` being stringified into the root `description`, same strip-and-stringify loop. Left out as a separate logical change.


lucachitayat (commented):
Self-review notes on the decisions a reviewer would most reasonably question.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -254,6 +254,16 @@ public static AITool AsAITool(this ToolUnion tool)
         "minItems",
     };
 
+    /// <summary>Gets the properties supported for schemas of the given JSON Schema type.</summary>
+    private static HashSet<string> GetSupportedSchemaProperties(string? type) =>
```
lucachitayat, line 258: Extracted verbatim from the inline `type switch` that used to live at the strip site — same arms, same fall-through to the base set — so it can be called once per member of a type array. No content change; the diff looks larger than the behaviour change is.

Takes `string?` so `types.FirstOrDefault()` on an empty set lands on the `_` arm, which keeps the typeless case a single expression instead of a third branch.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -289,18 +299,40 @@ public static AITool AsAITool(this ToolUnion tool)
...
-                    // Determine the schema type for type-specific handling.
-                    string? type =
-                        schemaObj.TryGetPropertyValue("type", out JsonNode? typeNode)
-                        && typeNode is JsonValue
-                            ? typeNode.GetValue<string>()
-                            : null;
+                    // Determine the schema type(s) for type-specific handling. A union type
+                    // ("type": ["string","null"], the shape System.Text.Json emits for
+                    // Nullable<T> and for a nullable reference type) is a JSON array rather
+                    // than a string, and carries the keywords of every member it names.
```
lucachitayat, line 308: This is the actual defect: the old sniff was `typeNode is JsonValue`, so a `JsonArray` produced `type == null` and the node took the base allowlist.

The array branch skips a non-string member instead of throwing. The old scalar path called `GetValue<string>()` unguarded, so `{"type": 3}` threw from inside the transform; both paths now use `TryGetValue`. Strictly a widening, and in line with the "don't throw on data the SDK doesn't recognize" note in `CLAUDE.md` — but it is a behaviour change beyond the headline fix, so flagging it rather than burying it.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -289,18 +299,40 @@ public static AITool AsAITool(this ToolUnion tool)
...
+                            types.Add(typeName);
+                        }
+                    }
 
                     List<KeyValuePair<string, string>>? removed = null;
 
                     // String format: only supported formats are kept.
                     if (
-                        type == "string"
+                        types.Contains("string")
```
lucachitayat: `type == "string"` becomes a membership test, so the unsupported-format check now fires on `["string","null"]` too. That is deliberate and is the half of the change that *keeps* stripping: `{"type":["string","null"],"format":"phone"}` is still stringified into `description`, exactly as the scalar `"string"` case is. A union node is treated on the same terms as its scalar counterpart in both directions. Same reasoning for `minItems` below.

**Both checks were also rewritten in the follow-up commit**, because widening them exposed two latent problems. The `format` check read its value with an unguarded `GetValue<string>()`, so a union node carrying a non-string `format` threw `InvalidOperationException` out of `GetResponseAsync` where it previously produced a valid schema; and both checks failed *open* on a malformed value, letting `"format": null` / `"minItems": "5"` / an out-of-int-range `minItems` through to the wire. They now keep the keyword only when its value is well formed and supported. `GetResponseAsync_ResponseFormatSchema_MalformedKeywordValuesAreStripped` pins all five cases, and the same throw on a scalar `"string"` node — which predates this PR — is fixed along with them.

src/Anthropic/AnthropicClientExtensions.cs
```diff
@@ -326,13 +358,21 @@ public static AITool AsAITool(this ToolUnion tool)
...
+                    // ["string","null"] keeps "format" exactly as "string" does.
+                    HashSet<string> supported;
+                    if (types.Count > 1)
                     {
-                        "object" => s_supportedObjectSchemaProperties,
-                        "string" => s_supportedStringSchemaProperties,
-                        "array" => s_supportedArraySchemaProperties,
-                        _ => s_supportedBaseSchemaProperties,
-                    };
+                        supported = new(s_supportedBaseSchemaProperties, StringComparer.Ordinal);
```
lucachitayat: The union-of-allowlists selection, and the main place I would welcome a second opinion.

This keeps the caller's `["string","null"]` verbatim and stops dropping keywords off it. The alternative — [anthropic-sdk-python#1892](https://github.com/anthropics/anthropic-sdk-python/pull/1892)'s expansion into `{"anyOf": [{"type": "string", "format": "date"}, {"type": "null"}]}` — produces a different but equally valid wire shape. I took the narrower route because `CLAUDE.md` says to map inputs as-is rather than rewrite what the caller built, and because the docs count `anyOf` and type arrays identically against the same 16-parameter limit, so the expansion buys nothing. Easy to switch if you'd rather have parity now; only the test expectations would change.

**Correcting an earlier version of this comment**, which claimed the scalar path "allocates nothing new" — that was wrong. As first written, a `HashSet` was allocated per node unconditionally, including for schemas containing no union types at all. That is fixed in the follow-up commit: `unionTypes` is now allocated only when `type` is an actual JSON array, so the scalar path allocates nothing it did not before. Worth knowing either way: `AIJsonSchemaTransformCache` keys a `ConditionalWeakTable` on *reference identity*, so it is once-per-instance, not once-per-type — building `ChatOptions.ResponseFormat` fresh per request (the idiomatic pattern, and what the SDK's own tests do) misses the cache every time.

src/Anthropic.Tests/AnthropicClientExtensionsTestsBase.cs
```diff
@@ -8565,6 +8565,143 @@ public async Task GetResponseAsync_ResponseFormatSchema_AllTransformationsApplie
         Assert.NotNull(response);
     }
 
+    /// <summary>
+    /// Validates that a union type ("type": ["string","null"], the shape System.Text.Json emits
+    /// for Nullable&lt;T&gt; and for a nullable reference type) keeps the keywords of every member
```
lucachitayat, line 8570: In the shared base, so it runs against both adapters.

The non-nullable `startsOn` control rides the same schema on purpose: it separates "nullability breaks `format`" from "the transform always strips `format`". Without it the test would still fail-before and pass-after, but it would not say *why*.

Members chosen to cover each allowlist that exists: `["string","null"]` for `format`, `["array","null"]` for `items`, `["object","null"]` for `properties`/`required`/`additionalProperties`. There is no number/integer allowlist to cover — `minimum`/`maximum`/`multipleOf` fall to the base set on every node, union or not, which matches the documented SDK-validated treatment and is unchanged here.

A second test, `…_MalformedKeywordValuesAreStripped`, was added in the follow-up commit for the malformed-value cases.

sophie-ant (approved):

sophie-ant:
Thanks!

Issue #255, opened 2026-09-01
MEAI adapter: HostedMcpServerTool requests are always rejected


johanjeppson:
## Environment

- `Anthropic` 12.44.0
- `Microsoft.Extensions.AI.Abstractions` 10.9.0
- .NET 10

## Repro

Any `HostedMcpServerTool` passed through the beta `AsIChatClient`:

```csharp
var client = new AnthropicClient(new ClientOptions { ApiKey = apiKey }).Beta.AsIChatClient();

var options = new ChatOptions
{
    ModelId = "claude-haiku-4-5",
    MaxOutputTokens = 1024,
    Tools = [new HostedMcpServerTool("microsoft_learn", "https://learn.microsoft.com/api/mcp")]
};

var response = await client.GetResponseAsync(
    [new ChatMessage(ChatRole.User, "What tools are available?")], options);
```

## Actual

```
Anthropic.Exceptions.AnthropicBadRequestException: Status Code: BadRequest
{"type":"error","error":{"type":"invalid_request_error","message":"MCP server 'mcp' is defined but not referenced by any `mcp_toolset` in `tools`."}}
```

Note the server is reported as `'mcp'` even though it was configured as `microsoft_learn`.

## Cause

Three issues in the `HostedMcpServerTool` branch of the beta tool mapping
(`AnthropicBetaClientExtensions`):

1. **Wrong name property.** The server definition is built with `Name = mcp.Name` — but
   `HostedMcpServerTool.Name` (the `AITool.Name` override) is hardcoded to the literal `"mcp"`
   in Microsoft.Extensions.AI 10.9; the configured server name is `ServerName`, which the
   mapping never reads. Every server serializes as `"name": "mcp"`, so multiple servers also
   collide.
2. **No `mcp_toolset` is emitted.** The mapping adds the server to `mcp_servers` and opts into
   the `mcp-client-2025-11-20` beta, but never adds a `BetaMcpToolset` to `tools`. That API
   shape requires every `mcp_servers` entry to be referenced by exactly one MCPToolset
   ([docs — validation rules](https://platform.claude.com/docs/en/agents-and-tools/mcp-connector#validation-rules)),
   so the request is rejected unconditionally.
3. **Deprecated allowlist shape.** `AllowedTools` is mapped to `tool_configuration` on the
   server definition, which is deprecated under `mcp-client-2025-11-20` — the allowlist now
   belongs in the toolset's `default_config`/`configs`
[...]

PR #267, opened 2026-09-17
fix(meai): map HostedMcpServerTool to MCP toolsets

z0rimo:
Fixes #255.

## Summary

Fix the beta MEAI `HostedMcpServerTool` mapping to use the current `mcp-client-2025-11-20` MCP connector request shape.

The existing adapter serializes `HostedMcpServerTool.Name`, whose value is the fixed tool-type identifier `"mcp"`, instead of the configured `ServerName`. As a result, distinct hosted MCP servers are serialized with the same wire-level name.

For example, the existing multiple-server request-shape test configures servers named `"server1"` and `"server2"`, but the previous mapping serializes both as:

```json
"name": "mcp"
```

The adapter also adds hosted MCP servers to `mcp_servers` without emitting the corresponding `mcp_toolset`.

As reported in #255, this produces an API rejection:

```text
MCP server 'mcp' is defined but not referenced by any `mcp_toolset` in `tools`.
```

The current [[MCP connector validation rules](https://platform.claude.com/docs/en/agents-and-tools/mcp-connector#validation-rules)](https://platform.claude.com/docs/en/agents-and-tools/mcp-connector#validation-rules) require every server in `mcp_servers` to be referenced by exactly one `MCPToolset`.

Finally, `AllowedTools` is currently mapped through the deprecated server-level `tool_configuration`; with `mcp-client-2025-11-20`, tool configuration belongs on the corresponding `mcp_toolset`.

## Changes

* use `HostedMcpServerTool.ServerName` for `mcp_servers[].name`
* emit one matching `BetaMcpToolset` for each hosted MCP server
* move `AllowedTools` mapping from the deprecated server-level `tool_configuration` to the toolset:

  * set `default_config.enabled = false`
  * explicitly enable each allowed tool through `configs.<tool>.enabled = true`
* preserve the existing `mcp-client-2025-11-20` beta header behavior
* preserve hosted MCP server/toolset ordering
* keep `null` and empty `AllowedTools` behavior unchanged

No automatic merging or de-duplication is performed for caller-supplied raw `BetaMcpToolset` values.

The stable adapter is unchanged because the stable request surface does not expose `McpServers`.

## Tests

Updated the existing beta MEAI request-shape tests for:

* a single hosted MCP server
* `AllowedTools`
* multiple hosted MCP servers
* hosted-tool beta-header handling
* preservation/de-duplication of existing beta headers

The multiple-server case also directly covers the `Name` → `ServerName` regression: two separately configured server names now remain distinct in the serialized `mcp_servers` array instead of both becoming `"mcp"`.

Validation:

* focused hosted MCP tests: 5 passed
* `./scripts/format`: passed
* `./scripts/build`: passed, 0 warnings / 0 errors
* `./scripts/lint`: passed
* full `net8.0` suite before the final test-only cleanup: 15,097 passed / 0 failed / 31 skipped
* `git diff --check`: passed
* `git show --check HEAD`: passed

The local Linux environment does not have Mono installed, so the `net472` leg of `./scripts/test` could not run locally. That target is covered by Windows CI.


feiiiiii5 (commented):
Reviewing this because it touches the same file as #270 and #271 and I already had the harness up. Everything below was checked against `main` at `c4a1b81e` and run, not read.

**The `Name` → `ServerName` half is a real fix, and the repo can prove it without any external API docs.**

`HostedMcpServerTool.Name` really is `"mcp"`. The NuGet XML doc for that property is just `<inheritdoc />`, so instead of reasoning from metadata I ran the base expectations. At `main`, six hosted-MCP assertions in `AnthropicClientBetaExtensionsTests` expect `"name": "mcp"` (lines 134, 199, 268, 273, 1562, 1634) and pass:

```
dotnet exec src/Anthropic.Tests/bin/Debug/net8.0/Anthropic.Tests.dll \
  -class Microsoft.Extensions.AI.Tests.AnthropicClientBetaExtensionsTests -method '*Hosted*'
Total: 36, Errors: 0, Failed: 0, Skipped: 0
```

The comparison is exact — `AnthropicClientExtensionsTestsBase.cs:10351-10354` is `Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expectedRequest), JsonNode.Parse(actualRequest)), …)` — so those passes are only possible if `Name` is literally `"mcp"`.

The clearest evidence is `GetResponseAsync_WithMultipleHostedMcpServerTools`: it builds two tools with distinct server names (lines 309-310, `"server1"` / `"server2"`), and the base expectation serializes **both** `mcp_servers` entries as `"name": "mcp"` (lines 268, 273). Two servers, one name. `AgentCreateParams.cs:72-75` states the contract for the Agents surface: "Names must be unique within the array. Every server must be referenced by an `mcp_toolset` in `tools`; unreferenced servers are rejected." On that reading the base behaviour cannot work for two or more MCP servers at all.

Suggestion: lead the PR description with the duplicate-name half. It is the part that is provable from inside the repo; the toolset-pairing half is not (below).

**Type-level details check out**, so nobody has to re-derive them:

- `createdTools` is `List<BetaToolUnion>` (`AnthropicBetaClientExtensions.cs:1362`), and `BetaToolUnion` has `public static implicit operator BetaToolUnion(BetaMcpToolset value) => new(value);` (`BetaToolUnion.cs:1656`), so adding the toolset to that list is legal and serializes through the union variant's own `Json`.
- `new BetaMcpToolset(mcp.ServerName)` does set `type`. The `(string)` ctor (`BetaMcpToolset.cs:149-153`) chains to the parameterless one (`:117-120`), which sets `Type = JsonSerializer.SerializeToElement("mcp_toolset")`. The `with` expression goes through the copy ctor, which copies `_rawData`, so `type` survives adding `DefaultConfig`/`Configs`.
- Neither `DefaultConfig` nor `Configs` is assigned null, so the CLAUDE.md "sending `null` where the property should be omitted" trap is avoided and the empty-`AllowedTools` case emits neither key.

**One claim I could not verify in-repo — and it is the one the new code comment states as fact.**

The comment says "The API requires each server to be referenced by exactly one `mcp_toolset`". Searching `main`, that pairing requirement is documented only on the **Agents** surface (`AgentCreateParams.cs:72-75`, and the same wording in `AgentUpdateParams.cs`), and there it says *every* server must be referenced — not *exactly one*. On the Messages surface `MessageCreateParams.cs:282-284` says only "MCP servers to be utilized in this request", `BetaRequestMcpServerUrlDefinition` has no doc comments, and there is no wire-shape test or example pairing `mcp_servers` with `mcp_toolset`. ("exactly one" does occur in the repo, but only about browser-state tabs and User-Agent headers — never about `mcp_toolset`.)

So the Messages-side requirement rests on the API docs or the rejection you saw, not on anything a reviewer can check here. The description says "requests using `HostedMcpServerTool` were rejected by the API" — could you paste the actual error response? That would settle both the shape and whether the contract is "exactly one" or "at least one", and it is the first thing a maintainer will ask for. If it is "at least one", the comment is worth softening so the SDK does not document a stronger contract than it can support.

**Two smaller things**

- This targets `next`. CLAUDE.md:10 says PRs target the default branch and that `next` belongs to release automation; 13 of the 14 most recently merged PRs here target `main` and none target `next`. Rebasing is cheap and changes nothing in this diff — `main` is currently `next` plus two release-metadata commits, and `git diff --name-only origin/next origin/main -- '*.cs'` is empty.
- The five rewritten expectations were the only thing pinning the old wire shape, so rewriting them is correct. It does mean no test in the diff fails before the source change and passes after. If you want the fix to be self-evidently load-bearing, the two-server case is the one to point at: on base it expects `"name": "mcp"` twice; after the change it expects `"server1"` and `"server2"`.

Not a maintainer and no stake in the design. Flagging the contract claim because it is the only part I could not check, and the duplicate-name evidence because it argues the case more strongly than the description currently does.

z0rimo:
> Reviewing this because it touches the same file as #270 and #271 and I already had the harness up. Everything below was checked against `main` at `c4a1b81e` and run, not read.
> 
> **The `Name` → `ServerName` half is a real fix, and the repo can prove it without any external API docs.**
> 
> `HostedMcpServerTool.Name` really is `"mcp"`. The NuGet XML doc for that property is just `<inheritdoc />`, so instead of reasoning from metadata I ran the base expectations. At `main`, six hosted-MCP assertions in `AnthropicClientBetaExtensionsTests` expect `"name": "mcp"` (lines 134, 199, 268, 273, 1562, 1634) and pass:
> 
> ```
> dotnet exec src/Anthropic.Tests/bin/Debug/net8.0/Anthropic.Tests.dll \
>   -class Microsoft.Extensions.AI.Tests.AnthropicClientBetaExtensionsTests -method '*Hosted*'
> Total: 36, Errors: 0, Failed: 0, Skipped: 0
> ```
> 
> The comparison is exact — `AnthropicClientExtensionsTestsBase.cs:10351-10354` is `Assert.True(JsonNode.DeepEquals(JsonNode.Parse(expectedRequest), JsonNode.Parse(actualRequest)), …)` — so those passes are only possible if `Name` is literally `"mcp"`.
> 
> The clearest evidence is `GetResponseAsync_WithMultipleHostedMcpServerTools`: it builds two tools with distinct server names (lines 309-310, `"server1"` / `"server2"`), and the base expectation serializes **both** `mcp_servers` entries as `"name": "mcp"` (lines 268, 273). Two servers, one name. `AgentCreateParams.cs:72-75` states the contract for the Agents surface: "Names must be unique within the array. Every server must be referenced by an `mcp_toolset` in `tools`; unreferenced servers are rejected." On that reading the base behaviour cannot work for two or more MCP servers at all.
> 
> Suggestion: lead the PR description with the duplicate-name half. It is the part that is provable from inside the repo; the toolset-pairing half is not (below).
> 
> **Type-level details check out**, so nobody has to re-derive them:
> 
> * `createdTools` is `List<BetaToolUnion>` (`AnthropicBetaClientExtensions.cs:1362`), and `BetaToolUnion` has `public static implicit operator BetaToolUnion(BetaMcpToolset value) => new(value);` (`BetaToolUnion.cs:1656`), so adding the toolset to that list is legal and serializes through the union variant's own `Json`.
> * `new BetaMcpToolset(mcp.ServerName)` does set `type`. The `(string)` ctor (`BetaMcpToolset.cs:149-153`) chains to the parameterless one (`:117-120`), which sets `Type = JsonSerializer.SerializeToElement("mcp_toolset")`. The `with` expression goes through the copy ctor, which copies `_rawData`, so `type` survives adding `DefaultConfig`/`Configs`.
> * Neither `DefaultConfig` nor `Configs` is assigned null, so the CLAUDE.md "sending `null` where the property should be omitted" trap is avoided and the empty-`AllowedTools` case emits neither key.
> 
> **One claim I could not verify in-repo — and it is the one the new code comment states as fact.**
> 
> The comment says "The API requires each server to be referenced by exactly one `mcp_toolset`". Searching `main`, that pairing requirement is documented only on the **Agents** surface (`AgentCreateParams.cs:72-75`, and the same wording in `AgentUpdateParams.cs`), and there it says _every_ server must be referenced — not _exactly one_. On the Messages surface `MessageCreateParams.cs:282-284` says only "MCP servers to be utilized in this request", `BetaRequestMcpServerUrlDefinition` has no doc comments, and there is no wire-shape test or example pairing `mcp_servers` with `mcp_toolset`. ("exactly one" does occur in the repo, but only about browser-state tabs and User-Agent headers — never about `mcp_toolset`.)
> 
> So the Messages-side requirement rests on the API docs or the rejection you saw, not on anything a reviewer can check here. The description says "requests using `HostedMcpServerTool` were rejected by the API" — could you paste the actual error response? That would settle both the shape and whether the contract is "exactly one" or "at least one", and it is the first thing a maintainer will ask for. If it is "at least one", the comment is worth softening so the SDK does not document a stronger contract than it can support.
> 
> **Two smaller things**
> 
> * This targets `next`. CLAUDE.md:10 says PRs target the default branch and that `next` belongs to release automation; 13 of the 14 most recently merged PRs here target `main` and none target `next`. Rebasing is cheap and changes nothing in this diff — `main` is currently `next` plus two release-metadata commits, and `git diff --name-only origin/next origin/main -- '*.cs'` is empty.
> * The five rewritten expectations were the only thing pinning the old wire shape, so rewriting them is correct. It does mean no test in the diff fails before the source change and passes after. If you want the fix to be self-evidently load-bearing, the two-server case is the one to point at: on base it expects `"name": "mcp"` twice; after the change it expects `"server1"` and `"server2"`.
> 
> Not a maintainer and no stake in the design. Flagging the contract claim because it is the only part I could not check, and the duplicate-name evidence because it argues the case more strongly than the description currently does.

Thanks for the thorough review — especially for checking the existing request-shape tests against `main`.

You're right about the base branch. This PR was originally opened from an older checkout where `CLAUDE.md` still directed PRs to `next`; the current guidance targets `main` and reserves `next` for release automation. I've rebased onto the current `main` and retargeted the PR accordingly.

For the MCP toolset requirement, the current Messages MCP connector documentation explicitly requires every server in `mcp_servers` to be referenced by exactly one `MCPToolset`, so I'm keeping that code comment and have added the documentation reference to the PR description.

I also clarified the API-rejection claim. I didn't personally make the live request; the actual 400 response is the one reported in #255:

`MCP server 'mcp' is defined but not referenced by any mcp_toolset in tools.`

I've attributed that observation to #255 rather than presenting it as a request I ran myself, and moved the duplicate-name issue forward in the description since, as you noted, that part is directly demonstrated by the existing two-server test.

I kept the implementation and test set otherwise unchanged; the existing multiple-server request-shape test directly captures the `Name` → `ServerName` regression.

PR #271, opened 2026-09-18
fix(meai): relabel ModelId at the fallback seam in beta streaming

feiiiiii5:
## What was wrong

In the beta MEAI streaming adapter a `fallback` content block was never handled, so two things went wrong at once.

A `fallback` block marks the seam where the model that was asked declines and a different model completes the turn. `GetStreamingResponseAsync` latches `modelID` once, from `message_start`:

```csharp
if (string.IsNullOrWhiteSpace(modelID))
{
    modelID = rawMessageStart.Message.Model;
}
```

and that variable is read in exactly two places, both when building a `ChatResponseUpdate` (`ModelId = modelID`). Since the `content_block_start` switch had no `case BetaFallbackBlock:` — and no `default:` either — the seam was invisible: `modelID` kept the declining model's id, so every update after the seam, including the terminal usage update, told the caller that the model which refused produced the output it was holding. The block itself was dropped as well; an unmatched block yields an update with empty `Contents`, so nothing reached `RawRepresentation` either.

The rest of the SDK treats this block as load-bearing, and `BetaMessageContentAggregator` states the intended semantics outright:

```csharp
// The final hop's fallback block names the model that served the response —
// keeps the aggregated message consistent with the relabeled non-streaming message.
if (mergedBlock.Value is BetaFallbackBlock fallbackBlock)
{
    model = fallbackBlock.To.Model;
}
```

- The non-streaming MEAI path reads `ModelId` from the response (`createResult.Model.Raw() ?? createParams.Model.Raw()`), which the API has already relabelled, and passes unmodelled blocks through `ContentBlockValueToAIContent`, whose `default:` is `new AIContent() { RawRepresentation = blockValue }`.
- `BetaToolRunner` treats the block as the seam that decides which tool calls may execute (`Streaming_FallbackSeam_OnlyPostSeamToolsExecuted`).
- The behaviour is already pinned by an existing test, `CreateStreamingAggregation_RelabelsModelFromFallbackBlock`, which streams a seam and then asserts both halves: `Assert.IsType<BetaFallbackBlock>(stream.Content[0].Value)` and `Assert.Equal(Messages::Model.ClaudeSonnet4_6, stream.Model.Value())`.

So streaming was the one path that neither relabelled nor surfaced it, and a MEAI consumer logging or routing by `ChatResponseUpdate.ModelId` — cost attribution, per-model evals, a retry policy keyed on model — would silently attribute post-seam output to the wrong model. This PR makes the MEAI streaming adapter satisfy the same contract that test already states for the aggregator.

## What changed

One `case` in the `content_block_start` switch:

```csharp
case BetaFallbackBlock fallback:
    // Streaming latches modelID at message_start, but a fallback block
    // means a different model served the rest of the turn. The
    // aggregator relabels the message from this block; without doing the
    // same here every later update names the model that declined.
    modelID = fallback.To.Model.Raw();
    contents.Add(
        ContentBlockValueToAIContent(
            contentBlockStart.ContentBlock.Value
        )
    );
    break;
```

The block is surfaced through the file's own pass-through helper rather than a hand-built `AIContent`, which is what the neighbouring group of blocks that arrive complete in `content_block_start` already does, and what CLAUDE.md asks for ("Not every block has deltas: some arrive complete in `content_block_start` … so default to passing whole blocks through"). Output is identical to `new AIContent() { RawRepresentation = fallback }` today, but it stays correct if that helper ever learns to model fallback blocks.

## One judgement call worth a second opinion

CLAUDE.md warns against throwing on data the SDK does not recognise, "especially in streaming and middleware code, where an exception tears down the caller's response". `ApiEnum<,>.Raw()` does throw `AnthropicInvalidDataException` when the raw value is not a `string`, so reading the model this way is not risk-free.

I kept `.Raw()` because the alternative is worse and the precedent is in this same switch: the `BetaServerToolUseBlock` arm already calls `serverToolUse.Name.Value()`, and CLAUDE.md itself notes `Value()` "is an undefined member for values the SDK doesn't know" — strictly more fragile than `Raw()`, which handles unknown members fine. `to.model` is a `required ApiEnum<string, Model>` and `BetaFallbackInfo.Validate()` is itself `this.Model.Raw();`, so the model type already treats `Raw()` as this field's contract; the non-streaming adapter reads its `ModelId` the same way. Guarding it here would mean a `try`/`catch` that decides what `ModelId` should be when the API sends a non-string model, which is a policy question rather than this bug. Happy to change it if you would rather the aggregator's approach (keep the `ApiEnum`, defer the read) were mirrored here instead.

## What gives confidence

New test `GetStreamingResponseAsync_FallbackBlock_RelabelsModelIdAndSurfacesTheBlock` in `AnthropicClientBetaExtensionsTests` (beta-only, so the derived class — the stable `ContentBlock` union has no fallback variant). The wire fixture uses the repo's canonical fallback shape, `trigger` included, as in `HistoryTrimmerTest`:

```
{"type":"fallback","from":{"model":"model-a"},"to":{"model":"model-b"},"trigger":{"type":"refusal","category":null}}
```

It asks for `model-a`, streams the seam plus a post-seam text block, then asserts two independent things: the update carrying `"Fallback Output"` has `ModelId == "model-b"`, and exactly one `AIContent` carries a `BetaFallbackBlock` whose `To.Model.Raw()` is `"model-b"`.

- On unmodified `main` at `c4a1b81e`, which is what this branch is rebased onto, the new test **fails**. Class run: `Total: 264, Failed: 1`.
- With the change: `Total: 264, Failed: 0`. The total is 264 in every run and `-list tests` confirms the new test is collected, so green cannot mean "never ran".
- Mutation checks, each expected to fail exactly the assertion that pins it:

| mutation | result |
|---|---|
| `modelID = fallback.To.Model.Raw();` → discard the value | new test fails (1 of 264) — `ModelId` is still `model-a` |
| drop the `contents.Add(...)` | new test fails (1 of 264) — `Assert.Single` finds no `BetaFallbackBlock` |

Both mutations fail alone, so neither assertion is passing on the other's account.

`./scripts/format` rc=0, `./scripts/lint` rc=0, build rc=0. Full `net8.0` suite: **15283 total, 37 failed, 41 skipped**, against a baseline measured on plain `main` at `c4a1b81e` of **15282 total, 37 failed, 41 skipped** — the +1 is this test. The failing set is identical before and after (compared as a set, not by count), and all 37 are pre-existing in this environment: every one is `AnthropicException: No workspace ID found. Set the WorkspaceId option or the ANTHROPIC_AWS_WORKSPACE_ID environment variable`, thrown while `AnthropicTestClientsAttribute.GetData` constructs an AWS client. `net472` was not run locally because it needs mono; CI runs it on Windows.

## Behaviour changes and limits

- Updates emitted *before* the seam still report the original model, which is correct — that output really did come from it. Only the seam and everything after it is relabelled.
- With several fallback blocks in one stream, the last one wins. That matches `BetaMessageContentAggregator`, which overwrites `model` on each block.
- The relabel changes `ChatResponseUpdate.ModelId` for callers of the beta streaming MEAI path. Anything that was reading the declining model's id out of a post-seam update was reading a wrong value; there is no other way to obtain the new model, since the block was previously dropped entirely.
- Only the beta adapter is touched. `git grep -l FallbackBlock` at `c4a1b81e` returns only `Models/Beta/Messages/*`, `Helpers/BetaMessageContentAggregator.cs` and the beta-side `Helpers/Fallbacks/*` machinery — there is no stable `FallbackBlock`, so CLAUDE.md's "a fix usually needs applying to every copy" has no second copy to apply to here.
- Relates to #270 (same file, same switch, different root cause: that one is about `mcp_tool_use` argument folding). The two are independent and the new tests sit far apart in the file, so either can merge first.

Prepared with AI assistance. All verification above was run locally against `main` at `c4a1b81e`.


feiiiiii5:
Cross-SDK precedent for the behaviour this PR adds, which I found while reading the Python SDK's streaming accumulator and think is worth having on the thread.

`anthropics/anthropic-sdk-python`, `src/anthropic/lib/streaming/_beta_messages.py:494-497` on `main` at `0af01906` (v1.7.0), inside `accumulate_event`'s `content_block_start` branch:

```python
if event.content_block.type == "fallback":
    # the final hop's fallback block names the model that served the response —
    # keeps the snapshot consistent with the relabeled non-streaming message
    current_snapshot.model = event.content_block.to.model
```

So the Python SDK relabels the model at the fallback seam **in the streaming accumulator**, at the same point in the event sequence this PR uses (`content_block_start`), for the same reason I gave above: the non-streaming path already relabels, and streaming was the one path that didn't, so the two disagreed about which model produced the message.

Two things this does and doesn't settle:

- It does answer "is relabelling during streaming the intended behaviour, or should streaming leave `ModelId` alone and let callers read the block?" — the answer from the sibling SDK is that streaming relabels, and that the accumulated message's `model` is meant to match the relabelled non-streaming message. That's the contract this PR is aligning the MEAI adapter with.
- It doesn't bear on the `.Raw()` judgement call I flagged in the description. The Python field is typed, so there's no equivalent of `ApiEnum<,>.Raw()` throwing on a non-string model; that question is specific to this codebase and still open.

Also worth noting for whoever reviews this: the same Python branch guards every optional `message_delta` field it accumulates except `stop_reason`/`stop_sequence`/`stop_details`, which I filed separately as anthropic-sdk-python#1940. Unrelated to this PR's diff — mentioning it only because it came out of the same read and this PR touches the equivalent C# accumulator.

No change requested by this comment; the branch is unchanged since `cc0e5907`.

